PR #107, opened 2025-01-01
feat: better type narrowing

### LINKED ISSUE #106: Better type narrowing of discord objects
There are a few places where functions, from discord.py, return wide types. Wider than is possible in the given state of the world. This causes type checkers to complain about various properties. Ideally, we should assert that what we claim should be happening, is happening.

A key example is in the channel selector. You filtered for only text channels. But the function, `bot.fetch_channel`, returns `GuildChannel | PrivateChannel | Thread`. We should add a check that narrows the type to only `GuildChannel | Thread`. EX `assert isinstance(channel, (discord.abc.GuildChannel, discord.Thread))`. 

<details><summary>Referenced code</summary>
<p>

https://github.com/ghostty-org/discord-bot/blob/c829d1bc7d6d86789a860f4e1b31a483856b1be0/app/components/move_message.py#L23-L41

</p>
</details> 

My question is if this should be handled with an `assert` or should things be gracefully returned. So instead of `assert`, returning some informative error message to the user (or in other cases skipping an invalid property).
If we are more matching the greater project that this "serves", asserts feel more acceptable (even though they are slightly less python-like)

CoderJoshDK:
Closes #106

The commands of this bot, should only ever be ran on specific channel types. I added more assertions that these channel types are correct. I only did this in areas of function ambiguity.

Another change that ended up making into here, is properly responding to the interaction for `select_channel`, when that ui had already been used. Not responding results in an API error. This change avoids that. 

<details><summary>assert vs cast</summary>
<p>

I ended up using asserts here over cast. This is in part because we are not passing in parameters that need to be narrowed, but rather dealing with wide return types. A cast is more lightweight but the code paths that are being hit, are controlled by "side effects." To avoid someone making a breaking "upstream" change, an `assert` should be used. 

</p>
</details>


app/utils.py
```diff
@@ -72,6 +72,7 @@ def _format_subtext(executor: discord.Member | None, msg_data: MessageData) -> s
     if reactions := msg_data.reactions.items():
         lines.append("   ".join(f"{emoji} x{count}" for emoji, count in reactions))
     if executor:
+        assert isinstance(msg_data.channel, discord.TextChannel)
```
trag1c, line 75: This fails when moving a message from a thread

app/components/move_message.py
```diff
@@ -46,6 +50,8 @@ async def select_channel(
             if isinstance(channel, discord.Thread)
             else (channel, discord.utils.MISSING)
         )
+        assert isinstance(webhook_channel, discord.TextChannel)
```
trag1c, line 53: This fails when trying to move a message to a forum post

trag1c (changes requested):

app/components/message_filter.py
```diff
@@ -48,6 +48,7 @@ async def check_message_filters(message: discord.Message) -> bool:
         await message.delete()
         if message.type not in REGULAR_MESSAGE_TYPES:
             continue
+        assert isinstance(message.channel, discord.TextChannel)
```
CoderJoshDK, line 51: This was left as a `TextChannel` only because it ensures that only full text channels are being filtered.

app/components/move_message.py
```diff
@@ -31,9 +32,13 @@ async def select_channel(
         self, interaction: discord.Interaction, sel: discord.ui.ChannelSelect
     ) -> None:
         if self._used:
+            await interaction.response.send_message(
+                "Channel selector was already used.", ephemeral=True
+            )
```
trag1c, line 37: This should be reverted as #115 will handle this part of code + it's out of scope for this PR anyway.

app/utils.py
```diff
@@ -20,6 +20,7 @@
 )
 
 Account = discord.User | discord.Member
+GuildTextChannels = discord.TextChannel | discord.ForumChannel | discord.Thread
```
trag1c, line 23: nit
```suggestion
GuildTextChannel = discord.TextChannel | discord.ForumChannel | discord.Thread
```

trag1c (changes requested):

app/components/move_message.py
```diff
@@ -34,6 +35,7 @@ async def select_channel(
             return
         self._used = True
         channel = await bot.fetch_channel(sel.values[0].id)
+        assert isinstance(channel, GuildTextChannel)
```
trag1c, line 38: `channel`'s type narrowing seems wrong here:
1. too broad: the channel selector doesn't allow choosing a `ForumChannel` (applies to `utils.py:76` too)
2. too narrow: it is later cast to `TextChannel`, while in reality it can also be a `Thread` (L61, L67)

app/components/move_message.py
```diff
@@ -46,6 +48,8 @@ async def select_channel(
             if isinstance(channel, discord.Thread)
             else (channel, discord.utils.MISSING)
         )
+        assert isinstance(webhook_channel, GuildTextChannel)
```
trag1c, line 51: I'm pretty sure `webhook_channel` cannot be a `Thread`

trag1c (changes requested):

trag1c (approved):

PR #110, opened 2025-01-02
feat: autoclose `#help` posts after 24h

### LINKED ISSUE #86: Auto-close posts with "Solved"/"Moved to GitHub" tag

trag1c:
Closes #86. This PR adds a task which looks through active posts in `#help` and closes the ones that:
* have any of the following tags: ✅ Solved, :octocat: Moved to GitHub
* have not had a new message in 24 hours

The task itself runs every hour.


app/components/autoclose.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import cast
+
+import discord
+from discord.ext import tasks
+
+from app.setup import bot, config
+from app.utils import snowflake_timestamp
+
+
+@tasks.loop(hours=24)
```
CoderJoshDK, line 12: (I have not properly read through all of the logic in the PR btw, just saw this one thing and though I would weigh in)

I find that something like this should be ran every hour. It really helps keep things cleaner. I also tend to see what other big servers do. Like dpy or python. And they have it at 1 hour https://github.com/Rapptz/RoboDanny/blob/39af9d71ffd5f099094a05352c18b987a1dc5d04/cogs/dpy.py#L225

trag1c: Thanks, will change.

app/utils.py
```diff
@@ -137,6 +140,12 @@ async def try_dm(account: Account, content: str) -> None:
         print(f"Failed to DM {account} with: {shorten(content, width=50)}")
 
 
+def snowflake_timestamp(snowflake: int) -> dt.datetime:
+    return dt.datetime.fromtimestamp(
+        ((snowflake >> 22) + DISCORD_EPOCH) / 1000, tz=dt.UTC
+    )
```
iceghost, line 146: nit: is there a reason why we don't use the official `discord.utils.snowflake_time` instead?

trag1c: Yes: I just missed it 😄 Thanks for the pointer

CoderJoshDK:
I wonder if it makes more sense to just have an auto close after X hours of inactivity, but no need to check the tags, because you create commands that add the tags and whatever other magic it needs to. For example, a solved command that adds the solved tag and archives the thread. And then a github command that ... well not sure how complex you want to make that one. 
If you don't want to go that route, or do but think it should be a different PR, then I think this one looks good. 

<details><summary>Github command idea</summary>
<p>

Some ideas of a github command include creating an issue for the user. Or having an option to tag it as moved since it was done by the user. I can see a few reasons why this isn't desired. But figured I would mention it.

</p>
</details>

iceghost (approved):

trag1c:
> I wonder if it makes more sense to just have an auto close after X hours of inactivity, but no need to check the tags, because you create commands that add the tags and whatever other magic it needs to.

Hm, I'm not convinced—it feels more sound to me to have the explicit/implicit divide (i.e. manual command/regular check).

PR #115, opened 2025-01-03
feat: streamline move message view

### LINKED ISSUE #111: Only allow one use of the "👻 Ghostping" button

### LINKED ISSUE #112: Streamline the move message interface (1 message only)
Right now, when you move a message, it sends you up to 3 ephemeral messages. One for selecting the channel. One for conformation of action, with an option to ghost ping someone. And if you do, a third ephemeral message that says you pinged them. However, this should ideally be just 1 message. As a side effect, having there is no need for the `_used` check, since the move UI would be replaced with the next view state.

There are some considerations / tradeoffs for making this a single ephemeral message. Currently, when moving a message, a message pops up saying that the bot is thinking. This is part of the defer action. However, by doing this, you replace the `interaction` reference to the channel selector. Meaning, you could no longer restrict this interaction to only 1 ephemeral message. The tradeoff means that there will be a slight delay where there is no "thinking" indicator and the view isn't updated to indicate that the action is completed / being worked on. The delay is super slight, but technically there is downtime. I think this is well worth it. But still figured it is worth mentioning.

The full streamlined interface looks like:
https://github.com/user-attachments/assets/3fff93b1-fd69-4c83-919b-c41970acafb0

If you are happy with this, I can create the PR. These changes also disable the button after use, so someone can't ping an author multiple times.

CoderJoshDK:
Closes #112
Closes #111 

This PR makes the move message flow, only use one view. Makes using it much cleaner.

Because we update the ghostping button and disable it after use, it prevents anyone from spamming pings to a user. This is why we respond to the interaction and don't defer it. By defer-ing it, someone could spam the button before it updates. That is also why the update happens before the actual ping. 

Also, if we defer and then try to edit the message on the followup, it will edit the defer and not the original message. Hence, the use of `edit_original_response`.


iceghost (approved):

trag1c (approved):
Looks good! I'll wait for #107 to merge first :)

PR #124, opened 2025-01-08
feat: send back the message content when filtered

### LINKED ISSUE #119: Filters should have a way to preserve what was removed
Generally, it isn't great when a message is deleted without a good way to retrieve that message. With the introduction of a log channel, I wonder if it would be wise to send the deleted message to the log channel. That (and/or) the DM to the offender, contains their original message. 

Imagine someone wanted to comment on some showcase but didn't know (or remember) that the channel deletes non attachment messages. Most of the time, this is simple enough to rewrite. But every now-n-again, it is a real pain.

This is a QoL request that isn't required. This can be ignored. I am unsure if the moderation bot on the main channel picks up on this and solves this problem already. If it does, it still might be worth thinking about sending the original message in the DM to the user. But how that message is sent, should be similar to the resolution of #118.

trag1c:
Closes #119

If the deleted message had any content, the bot will send it back to the user with the notification (it sends 3 messages so it's easier to use the "Copy Text" thing, but 2 of them are silent so the sound plays just once).

The behavior for no content is unchanged (the message below the red line wasn't cropped 😉).

### DIFF
--- a/app/components/message_filter.py
+++ b/app/components/message_filter.py
@@ -1,5 +1,6 @@
 import re
 from collections.abc import Callable
+from io import BytesIO
 from typing import NamedTuple
 
 import discord
@@ -12,7 +13,8 @@ URL_REGEX = re.compile(
 )
 MESSAGE_DELETION_TEMPLATE = (
     "Hey! Your message in {} was deleted because it did not contain {}."
-    " Make sure to include {}, and respond in threads."
+    " Make sure to include {}, and respond in threads.\n"
+    "Here's the message you tried to send:\n\n"
 )
 REGULAR_MESSAGE_TYPES = frozenset(
     {discord.MessageType.default, discord.MessageType.reply}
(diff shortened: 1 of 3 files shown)


00-kat (approved):

app/components/message_filter.py
```diff
@@ -45,15 +51,38 @@ async def check_message_filters(message: discord.Message) -> bool:
...
+                file=discord.File(
+                    BytesIO(message.content.encode()), filename="content.md"
+                ),
+            )
+        elif content_size > 0:
+            await try_dm(message.author, message.content)
+
+        if 0 < content_size <= 2000:
+            await try_dm(message.author, COPY_TEXT_HINT, silent=True)
+
```
iceghost, line 86: Can the logic for long message goes into `try_dm` itself? Feels like we have to do that every time `try_dm` is used anyway

trag1c: Other than here, this kind of check is only ever used in the moving messages feature where `try_dm` isn't used. But I agree this can probably be abstracted away for reuse 
and plugged into `try_dm` as a built-in safeguard 👍

iceghost (approved):
I have not tested this with a bot, but lgtm aside from my prior comment.

app/components/message_filter.py
iceghost: Alrighty. I felt like bringing this up since I have seen similar snippets on 2000 in some other pre. Looks like you already have a plan though :D

PR #129, opened 2025-01-15
build: replace `pygithub` with `githubkit`

trag1c:
Draft, trying to figure out how to make Nix like this change


Uzaaft (approved):
With #132  merged into this we can :shipit:

PR #132, opened 2025-01-16
perf: improve entity fetching flow

trag1c:
Built on top of #129. The wait time decreases significantly when multiple entities are mentioned for the first time in a message:
image


Uzaaft (approved):
:shipit: Blazingly fast

PR #151, opened 2025-02-09
Handle long help post titles gracefully

00-kat:
Before this change, if the new title was longer than 100 characters, an exception would be raised.

This is *very* hacky and probably the worst code I've ever PR'd.  Perhaps "Unable to change the post title." should be returned without all that effort to show the error from the Discord API; I can change the PR to do so if requested, though I would prefer to have the nicer message so the user of the command actually knows what happened.

### DIFF
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -9,6 +9,11 @@ from app.components.entity_mentions import entity_message
 from app.setup import bot, config
 from app.utils import is_dm, is_helper, is_mod
 
+# https://discord.com/developers/docs/topics/opcodes-and-status-codes#http-http-response-codes
+INVALID_REQUEST_DATA = 400
+# https://discord.com/developers/docs/topics/opcodes-and-status-codes#json
+INVALID_FORM_BODY = 50035
+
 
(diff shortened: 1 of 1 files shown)


app/components/close_help_post.py
```diff
@@ -123,15 +128,73 @@ async def close_post(
...
+                "message"
+            ]
+        except Exception:
+            try:
+                # Parsing something of the form:
+                # "Invalid Form Body\nIn name: Must be 100 or fewer in length."
+                returned_error = (
+                    str(e).splitlines()[1].removeprefix("In name: ").strip()
+                )
+            except IndexError:
```
00-kat, line 178: Should this line instead be the following?
```suggestion
            # While only IndexError is likely to be thrown, don't make
            # assumptions on the data.
            except Exception:
```

I think the only other exception that *can* be thrown is if `str()` fails, but I'm pretty sure that `__str__()` is a requirement for `Exception` anyway.

00-kat: Based on the conversation on Discord I'm going to leave it as-is.

PR #145, opened 2025-02-07
feat: add Zig syntax highlighting

trag1c:
Closes #137. This relies on my https://github.com/trag1c/zig-codeblocks project.

Still left to do:
- [x] **Highlighting**
  - [x] Stabilize the core mechanism → https://github.com/trag1c/zig-codeblocks/issues/1
  - [x] https://github.com/trag1c/zig-codeblocks/issues/3 so adjustments can be done in this repo
  - [x] Fix https://github.com/trag1c/zig-codeblocks/issues/5
- [x] **Discord**
  - [x] Split into multiple messages when too large
  - [x] React to edits/deletions
  - [x] Add an option to freeze the message (so it stops reacting)
  - [x] Add a "Replace my message" option (with a confirmation prompt)
  - [x] **Files**
    - [x] Highlight files too
    - [x] Send small files back as regular code blocks
    - [x] Simplify attachment processing → https://github.com/trag1c/zig-codeblocks/issues/6
- [x] ~~Fix the Nix build fail (@Uzaaft's on it)~~ Nix is no more (#161)

PR #154, opened 2025-02-15
improve utils; fix zig-codeblocks message unlinking

trag1c:
Built on top of #145.

* removed unused utils
* made utils a package
* added a common message linker to replace `entity_mentions` and `zig_codeblocks`'s custom separate linker
* removed an unused field in `docs.py/Entry`
* the autoclose task should no longer fail when the `on_ready` event is triggered after a reconnection


00-kat (approved):

00-kat (approved):

00-kat (approved):
GitHub doesn't want me to review it, but I guess I'll add a review anyway just in case™.

PR #156, opened 2025-02-16
Sanitize markdown in GitHub entity titles

### LINKED ISSUE #149: Markdown in GitHub entity titles are not sanitised
Some screenshots of what this causes:
Figure 1

Figure 2

A search for `00-kat/test` in the Ghostty bot testing Discord server reveals quite a few too, some of which are shown below.  Not all of them are related to Markdown sanitisation but rather general sanitisation of Discord features embedded with special syntax in text messages.
Figure 3
Figure 4
This one required `<@user id>`, very unlikely apart from intentional abuse, so should likely not be handled.  Included for completeness.
Figure 5
Embeds could be very large, this one was luckily small.
Figure 6
Also contrived.
Figure 7
Not as contrived since GitHub also has :emojiname: syntax, and someone who doesn't know GitHub doesn't parse titles may try using them.  Though one could argue this is preferable.
Figure 8
Also contrived, but people have a reason to abuse this as it may aid in causing harm.
Figure 9

00-kat:
Closes #149.

### DIFF
--- a/app/components/entity_mentions/fmt.py
+++ b/app/components/entity_mentions/fmt.py
@@ -8,10 +8,11 @@ from githubkit.versions.latest.models import Issue, PullRequest
 
 from app.components.entity_mentions.resolution import resolve_repo_signatures
 from app.setup import bot, config
+from app.utils import escape_special
 
 from .cache import Entity, EntityKind, entity_cache
 
-ENTITY_TEMPLATE = "**{kind} [#{entity.number}](<{entity.html_url}>):** {entity.title}"
+ENTITY_TEMPLATE = "**{kind} [#{entity.number}](<{entity.html_url}>):** {title}"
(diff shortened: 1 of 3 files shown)


app/components/entity_mentions/fmt.py
```diff
@@ -43,7 +44,8 @@ async def load_emojis() -> None:
 
 
 def _format_mention(entity: Entity, kind: EntityKind) -> str:
-    headline = ENTITY_TEMPLATE.format(kind=kind, entity=entity)
+    title = escape_special(entity.title)
```
00-kat, line 47: Not sure if I should inline this; I liked the formatting with the new `title` variable more, and it's more consistent with the rest of them, so I kept it this way.  Let me know if you want it changed.

PR #157, opened 2025-02-17
Move Nitro emojis

00-kat:


app/utils/webhooks.py
```diff
@@ -6,6 +8,20 @@
 
 GuildTextChannel = discord.TextChannel | discord.Thread
 
+_EMOJI_REGEX = re.compile(r"<(a?):(\w+):(\d+)>", re.ASCII)
+
+
+def _convert_nitro_emojis(s: str) -> str:
```
00-kat, line 14: Not sure where to put it.  It's unlikely to be used outside of the `move_message_via_webhook()` function, but it also doesn't feel right leaving it in `webhook.py`.  I can't put it in `__init__.py` as that would cause a circular dependency.

app/utils/webhooks.py
```diff
@@ -65,12 +85,17 @@ async def move_message_via_webhook(
 
 
 def format_or_file(
-    message: str, *, template: str | None = None
+    message: str,
+    *,
+    template: str | None = None,
+    filter: Callable[[str], str] | None = None,
```
00-kat, line 91: Not sure if the convention in this repo is to suffix these with an underscore.

00-kat: I guess Ruff's check `A001` makes the answer to this obvious.

PR #160, opened 2025-02-18
Update the documentation for preparing the `.env` file

00-kat:


README.md
```diff
@@ -112,10 +117,14 @@ permission):
 
 Create a `.env` file in the root of the project based on `.env.example`.
 Below are explanations for each variable:
+* `BOT_ACCEPT_INVITE_URL`: a URL to visit to accept the Ghostty invite
```
00-kat, line 120: Assuming one copies `.env.example` in verbatim before editing these, we could omit this one were that to be desired.

trag1c: I think it's safer to keep it in.

trag1c (approved):
Thanks!

PR #163, opened 2025-02-24
Handle long help post titles gracefully

00-kat:
Continuation of #151, thank you GitHub.

Before this change, if the new title was longer than 100 characters, an exception would be raised.

This is *very* hacky and probably the worst code I've ever PR'd.  Perhaps "Unable to change the post title." should be returned without all that effort to show the error from the Discord API; I can change the PR to do so if requested, though I would prefer to have the nicer message so the user of the command actually knows what happened.

### DIFF
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -189,10 +189,8 @@ async def close_post(
 
         delim = ";" if ":" in title_prefix else ":"
         title_prefix = title_prefix.strip("[]").lower()
-        if additional_reply:
-            additional_reply = f"Closed{delim} {title_prefix}.\n{additional_reply}"
-        else:
-            additional_reply = f"Closed{delim} {title_prefix}."
+        additional_reply = f"\n{additional_reply}" if additional_reply else ""
+        additional_reply = f"Closed{delim} {title_prefix}.{additional_reply}"
 
     if additional_reply:
         await post.send(additional_reply)
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -167,7 +167,7 @@ async def close_post(
         delim = ";" if ":" in title_prefix else ":"
         title_prefix = title_prefix.strip("[]").lower()
         additional_reply = f"\n{additional_reply}" if additional_reply else ""
-        additional_reply = f"Closed{delim} {title_prefix}.{additional_reply}"
+        additional_reply = f"**Closed{delim} {title_prefix}**.{additional_reply}"
 
     if additional_reply:
         await post.send(additional_reply)
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -151,32 +151,16 @@ async def close_post(
         #     of Error Messages provided by an errors object.
         # in https://discord.com/developers/docs/topics/opcodes-and-status-codes#json
         # Both approaches are going to be tried... here be dragons.
-        try:
-            # The format of e._errors that is being parsed below is
-            # {
-            #     "name": {
-            #         "_errors": [
-            #             {
-            #                 "code": "BASE_TYPE_MAX_LENGTH",
-            #                 "message": "Must be 100 or fewer in length.",
-            #             }
-            #         ]
-            #     }
-            # }
-            # This tries to make as few assumptions of the keys of the
-            # dictionary as possible, as the field is private.
-            returned_error = next(iter(next(iter(e._errors.values())).values()))[0][  # noqa: SLF001
-                "message"
-            ]
-        except (StopIteration, IndexError):
-            try:
+        returned_error = ""
+        match e._errors, str(e).splitlines():  # noqa: SLF001
+            case {"name": {"_errors": [{"code": _, "message": msg}]}}, _:
+                returned_error = msg
+            case _, [_, snd]:
                 # Parsing something of the form:
                 # "Invalid Form Body\nIn name: Must be 100 or fewer in length."
-                returned_error = (
-                    str(e).splitlines()[1].removeprefix("In name: ").strip()
-                )
-            except IndexError:
-                returned_error = None
+                # where the second line is in `snd`.
+                returned_error = snd.removeprefix("In name: ").strip()
+
         if not returned_error:
             followup = "Unable to change the post title."
         else:
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -165,10 +165,10 @@ async def close_post(
             # }
             # This tries to make as few assumptions of the keys of the
             # dictionary as possible, as the field is private.
-            returned_error = next(iter(next(iter(e._errors.values())).values()))[0][
+            returned_error = next(iter(next(iter(e._errors.values())).values()))[0][  # noqa: SLF001
                 "message"
             ]
-        except Exception:
+        except (StopIteration, IndexError):
             try:
                 # Parsing something of the form:
                 # "Invalid Form Body\nIn name: Must be 100 or fewer in length."
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -9,6 +9,11 @@ from app.components.entity_mentions import entity_message
 from app.setup import bot, config
 from app.utils import is_dm, is_helper, is_mod
 
+# https://discord.com/developers/docs/topics/opcodes-and-status-codes#http-http-response-codes
+INVALID_REQUEST_DATA = 400
+# https://discord.com/developers/docs/topics/opcodes-and-status-codes#json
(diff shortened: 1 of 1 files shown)


app/components/close_help_post.py
```diff
@@ -123,15 +128,73 @@ async def close_post(
...
+                returned_error = None
+        if not returned_error:
+            followup = "Unable to change the post title."
+        else:
+            # Try to fix up some grammar and punctuation.
+            returned_error = returned_error[0].lower() + returned_error[1:]
+            if returned_error[-1] not in ".?!":
+                returned_error += "."
+            returned_error = returned_error.replace("must be", "must have been")
+            followup = f"Unable to change the post title as it {returned_error}"
```
trag1c, line 188: How many different error message can there be for this case?

00-kat: I have no idea, Discord doesn't document that.  They only give a generic description of the error code.

If I were to guess:
- Post title too long – returned to us.
- Invalid Unicode – I have no clue.
- Forbidden characters (lookalikes? unsafe Unicode?) – Discord probably sanitises it.

app/components/close_help_post.py
```diff
@@ -123,15 +128,73 @@ async def close_post(
...
+            ]
+        except (StopIteration, IndexError):
+            try:
+                # Parsing something of the form:
+                # "Invalid Form Body\nIn name: Must be 100 or fewer in length."
+                returned_error = (
+                    str(e).splitlines()[1].removeprefix("In name: ").strip()
+                )
+            except IndexError:
+                returned_error = None
```
trag1c, line 179: This can be much cleaner with pattern matching:
```suggestion
        match e._errors, str(e).splitlines():
            case {"name": {"_errors": [{"code": _, "message": msg}]}}, _:
                returned_error = msg
            case _, [_, snd]:
                # Parsing something of the form:
                # "Invalid Form Body\nIn name: Must be 100 or fewer in length."
                returned_error = snd.removeprefix("In name: ").strip()
            case _:
                returned_error = None
```

app/components/close_help_post.py
trag1c, line 188: I'd say, since we know what the original `returned_error` will generally be, maybe this should be hard-coded instead? So something like this:
```suggestion
        if returned_error == TOO_LONG_ERROR:
            followup = "Unable to change the post title because it would become >100 characters long."
        else:
            followup = "Unable to change the post title."
```
or
```suggestion
        if returned_error == TOO_LONG_ERROR:
            reason = " because it would become >100 characters long"
        else:
            reason = ""
        followup = f"Unable to change the post title{reason}."
```

app/components/close_help_post.py
```diff
@@ -123,15 +128,73 @@ async def close_post(
...
+        else:
+            # Try to fix up some grammar and punctuation.
+            returned_error = returned_error[0].lower() + returned_error[1:]
+            if returned_error[-1] not in ".?!":
+                returned_error += "."
+            returned_error = returned_error.replace("must be", "must have been")
+            followup = f"Unable to change the post title as it {returned_error}"
+
+        delim = ";" if ":" in title_prefix else ":"
+        title_prefix = title_prefix.removeprefix("[").removesuffix("]").lower()
```
trag1c, line 191: Subjective (and generally not functionally equivalent), but:
```suggestion
        title_prefix = title_prefix.strip("[]").lower()
```

app/components/close_help_post.py
```diff
@@ -123,15 +128,73 @@ async def close_post(
...
+        if additional_reply:
+            additional_reply = f"Closed{delim} {title_prefix}.\n{additional_reply}"
+        else:
+            additional_reply = f"Closed{delim} {title_prefix}."
```
trag1c, line 195: Subjective again:
```suggestion
        additional_reply = f"Closed{delim} {title_prefix}." + (
            f"\n{additional_reply}" if additional_reply else ""
        )
```

trag1c (changes requested):

app/components/close_help_post.py
00-kat: > `returned_error == TOO_LONG_ERROR`

Not sure where I can get a `TOO_LONG_ERROR` from, but I could do a `if returned_error:` instead similar to what I had originally done.

app/components/close_help_post.py
00-kat: Hm.  I was thinking of doing a ternary conditional originally, but I'm not a fan of those extra brackets added.  I do agree that there is far too much duplication of strings there, maybe I'll do something like this instead:
```suggestion
        additional_reply = f"\n{additional_reply}" if additional_reply else ""
        additional_reply = f"Closed{delim} {title_prefix}.{additional_reply}"
```

app/components/close_help_post.py
trag1c: My suggestion assumed you'd throw a
```py
TOO_LONG_ERROR = "Must be 100 or fewer in length."
```
in there, I should've been more clear. :sweat_smile:

Also, I think it would be handy to forcefully crash this function if there's a different error than just "name too long". It would help with discovery both locally and on Sentry, as right now it just gets silenced with a generic followup message.

00-kat: > My suggestion assumed you'd throw a […] in there, I should've been more clear. 😅

I'd personally prefer to do something like `returned_error.contains("or fewer in length")`, though I agree that harms readability more than having the whole string there.

> I think it would be handy to forcefully crash this function if there's a different error than just "name too long"

Definitely, but I also don't trust discord to keep their strings straight[^1]…  That is, if they change the limit or something, our check is going to fail.  Though I'm fine with doing that and updating this code when we get this issue again on Sentry.

[^1]: Their docs specifically say this is an additional message, not something to be parsed.  So far I've only parsed the extra stuff discord.py adds on around Discord's message, as well as some best-effort string replacements.

00-kat: For now I'm going to rethrow if it's not a name that's too long, and do a `"or fewer in length" in returned_error`, let me know your opinion and I'll switch it out to whatever we decide on.

00-kat: Oh also—post title is empty.  But that can't be encountered by this code so that one doesn't matter.

trag1c: > For now I'm going to [...] do a `"or fewer in length" in returned_error`

Fine by me

app/components/close_help_post.py
00-kat: Hmm, I just realised that pattern matching results in the exact structure of the private field being hardcoded in the code too, while my original thing tried to assume as little as possible about the private field.  If we're going to go this route, I may as well just directly access the element itself.

app/components/close_help_post.py
00-kat: Done.

app/components/close_help_post.py
00-kat: Well that cleaned up the code a whole bunch… it now looks completely different from the original PR.

trag1c: The thing with direct access is that it's not safeguarded. E.g. if the name `_errors` changes, this code simply won't match and will try the next case, while direct access will crash the entire thing.

00-kat: I brought back the `except IndexError` though.

00-kat: Actually I had handled the wrong errors there, fixed now.

00-kat: My fix for the wrong errors was wrong too, fixed yet again.

00-kat: Though it could be argued that my list of exceptions isn't exhaustive, and pattern matching would be the Ruff-friendly™ way to say `except Exception: pass`…

app/components/close_help_post.py
00-kat: Guess I have the answer for invalid Unicode…  it consistently results in Discord rate-limiting me.  That's… not good.

app/components/close_help_post.py
```diff
@@ -123,15 +128,48 @@ async def close_post(
         return
 
     await interaction.response.defer(ephemeral=True)
+    followup = "Post closed."
```
trag1c, line 131: I found this being defined so early a bit confusing. I think any of these places is fine:
* right before the `try` block
* right after the `await` in `try` (at that point we know it was successful)
* in an `else` block (i.e. try/except/else)

trag1c (commented):
One nit, LGTM other than that.

app/components/close_help_post.py
00-kat: Done, I did the second suggestion.

trag1c (approved):

Issue #155, opened 2025-02-16
Bulk-moving messages


00-kat:
This is quite similar to the fabled #18; as Ghostty is no longer in a private beta, there isn't much need to *create* the threads themselves anymore, but mostly just move them.

I've not figured out the exact details yet; I'll edit this once I've done that.

PR #165, opened 2025-02-24
Ignore polls that have ended

00-kat:
Continuation of #159, thank you GitHub.

Closes section 6.3 of #155.

### DIFF
--- a/app/utils/webhooks.py
+++ b/app/utils/webhooks.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import datetime as dt
 import re
 from io import BytesIO
 from typing import TYPE_CHECKING
@@ -120,9 +121,22 @@ async def move_message_via_webhook(
         msg_data.attachments.append(file)
         content += "\n-# (content attached)"
 
+    # Discord does not like negative poll durations. Polls created by a Webhook
+    # cannot be ended manually, so simply discard polls which have ended.
+    if message.poll is None:
+        poll = discord.utils.MISSING
+    elif (
+        message.poll.expires_at is None
+        or dt.datetime.now(tz=dt.UTC) >= message.poll.expires_at
+    ):
+        content += "\n-# (unable to attach closed poll)"
+        poll = discord.utils.MISSING
+    else:
+        poll = message.poll
+
     msg = await webhook.send(
         content=content,
-        poll=message.poll or discord.utils.MISSING,
+        poll=poll,
         username=message.author.display_name,
         avatar_url=message.author.display_avatar.url,
         allowed_mentions=discord.AllowedMentions.none(),


00-kat:
For those wondering what those force-pushes are:
- Past me was stupid and did some random division nonsense.
- Fixing that made me realise past me was extra stupid.
- I then though past me was blabbing nonsense into a PR and went to correct it, then a little while later realised past me is actually smarter than current me and reverted my entire rewrite of this tiny PR.

app/utils/webhooks.py
```diff
@@ -89,7 +89,14 @@ async def move_message_via_webhook(
...
         content=content,
-        poll=message.poll or discord.utils.MISSING,
+        poll=(
+            message.poll
+            # Discord does not like negative poll durations. Polls created by
+            # a Webhook cannot be ended manually, so simply discard polls which
+            # have ended.
+            if message.poll is not None and message.poll.duration.total_seconds() > 0
+            else discord.utils.MISSING
+        ),
```
trag1c, line 99: As far as I can tell, `poll.duration` only shows the originally set duration rather than "time left". I have a hunch we want `poll.expires_at` here.

00-kat: I'm not sure.  That was what I thought too, and discord.py doesn't document `duration` all that well (“The duration of the poll.”), but based on my experimentation it seemed to be the time left until it expired…  It's a `timedelta` too.  I'll check the discord.py source code as well in a minute.

00-kat: In `Poll._from_data`:
> ```py
> duration = datetime.timedelta(hours=round((expiry - message.created_at).total_seconds() / 3600))
> ```
Looks like it's the time left until it expires.

trag1c: I also experimented with this and it always said `1:00:00` for me when I requested the `poll.duration` (even after a restart and a manual message fetch after some time).

00-kat: Likely because it's rounder to the nearest hour.

trag1c: Okay I checked again. Made a poll last for 8 hours and checked it's `.duration` after 5 hours (showing up as 3h left on Discord) and it still said `8:00:00`. I'm really not convinced `poll.duration` is the way to go here.

trag1c: Looking at that `duration` definition now, it feels pretty clear that it's not supposed to be a changing property (since it's likely just subtracting constant timestamps). Perhaps we can flip this around and have our condition be something like `now - (created_at + duration) ≥ 1h`?

00-kat: > Made a poll last for 8 hours and checked its `.duration` after 5 hours (showing up as 3h left on Discord)

Was your `message.poll` from when the message was created?  When moving a message with a poll that has ended, `.duration` appears to be `0:00:00`:
```python-repl
>>> print(message.poll)
<Poll duration=0:00:00 question="e" answers=[<PollAnswer id=1 media=<PollMedia text='e' emoji=None>>, <PollAnswer id=2 media=<PollMedia text='e' emoji=None>>]>
```
Likely because the `message` is fetched when a user attempts to move a message, and hence `.duration` is 0 if one moves a message with a closed poll.  I feel like this would break down if a poll closes in between the time when somebody starts moving a message and when `move_message_via_webhook()` is called, so I'm going to stop using `.duration` anyway.

00-kat: Fixed this.

> I feel like this would break down if a poll closes in between the time when somebody starts moving a message and when `move_message_via_webhook()` is called, so I'm going to stop using `.duration` anyway.

But now that I thought about this, turns out that it will un-end a poll if this happens.  It's *so unlikely* that this will ever happen in practice, so I'm not particularly interested in fixing this.

app/utils/webhooks.py
```diff
@@ -120,9 +121,20 @@ async def move_message_via_webhook(
         msg_data.attachments.append(file)
         content += "\n-# (content attached)"
 
+    # Discord does not like negative poll durations. Polls created by a Webhook
+    # cannot be ended manually, so simply discard polls which have ended.
+    if (
```
00-kat, line 126: It looks really odd if this is a ternary, since it's a pretty long condition now, so I decided to pull it out into its own if statement.

app/utils/webhooks.py
trag1c: > Was your `message.poll` from when the message was created?

Nope, I mentioned that I checked it with a fresh instance (so I just wrote some code to fetch the exact message from ID and did a completely new bot run).

> When moving a message with a poll that has ended, `.duration` appears to be `0:00:00`

Makes sense looking at that code earlier (since it's `expiry - message_created.at` and a moved message has a new timestamp).

00-kat: > a moved message has a new timestamp

That was from the original message that was to be moved.

trag1c (commented):
Okay, this seems to work correctly now (assuming it creating a new poll with the original duration if not ended is correct).

My only concern now is: _should we even move a message with an ended poll?_ I'm not aware what a poll message could contain other than reactions, so for a ended poll with no reactions it'd just end up being an empty `Moved from <…>`  message.

00-kat:
Not sure if it's possible without a modded client, but it is in theory possible to edit a poll to add text to it.  But my reasoning for sending an empty message is so that a user can actually tell that a message of theirs had been moved and not just deleted.  Perhaps a placeholder message with an error would be more useful though.

00-kat:
> assuming it creating a new poll with the original duration if not ended is correct

Also, yes—this is intended behaviour because it's not possible to add back poll votes though webhooks, hence necessitating that interval for voters to add their votes back.

trag1c:
> Perhaps a placeholder message with an error would be more useful though.

Yeah let's do that.

app/utils/webhooks.py
```diff
@@ -120,9 +121,21 @@ async def move_message_via_webhook(
...
+        message.poll is None
+        or message.poll.expires_at is None
+        or dt.datetime.now(tz=dt.UTC) >= message.poll.expires_at
+    ):
+        content += "\n-# (Unable to attach poll that ended)"
```
00-kat, line 131: Not sure how I feel about adding “that ended”; I'd like to hear your opinion on that part of this.  Also, I've tried to mimic the other extra context above, and am hence unsure if I should leave the U capitalised.

trag1c: Love some bikeshedding; looks like Discord just says the polls are closed. Additional notes seem to be in lowercase (e.g. `(content attached)`, `(skipped n large attachment(s))`) so I'll go with `(unable to attach a closed poll)` 👍

trag1c, line 131: This looks wrong—the note will be added _every_ time we don't add a poll, so even when there wasn't one in the original message.

trag1c (changes requested):

app/utils/webhooks.py
00-kat: Oops, overlooked that.  Fixed now.

trag1c (approved):

PR #166, opened 2025-02-24
Move nitro emojis

00-kat:
Continuation of #157, thank you GitHub.

Closes the first part of #155.

### DIFF
--- a/app/utils/webhooks.py
+++ b/app/utils/webhooks.py
@@ -1,10 +1,31 @@
+from __future__ import annotations
+
+import re
 from io import BytesIO
+from typing import TYPE_CHECKING
 
 import discord
 
 from app.utils.message_data import MessageData, scrape_message_data
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
 GuildTextChannel = discord.TextChannel | discord.Thread
+
+_EMOJI_REGEX = re.compile(r"<(a?):(\w+):(\d+)>", re.ASCII)
+
+
+def _convert_nitro_emojis(content: str) -> str:
+    def r(match):
+        flag = match.group(1)
+        ext = "gif" if flag else "webp"
+        tag = "&animated=true" if flag else ""
+        name = match.group(2)
+        id_ = match.group(3)
+        return f"[{name}](https://cdn.discordapp.com/emojis/{id_}.{ext}?size=48{tag}&name={name})"
+
+    return _EMOJI_REGEX.sub(r, content)
 
 
 def _format_subtext(executor: discord.Member | None, msg_data: MessageData) -> str:
@@ -44,7 +65,11 @@ async def move_message_via_webhook(
     msg_data = await scrape_message_data(message)
 
     subtext = _format_subtext(executor, msg_data)
-    content, file = format_or_file(msg_data.content, template=f"{{}}{subtext}")
+    content, file = format_or_file(
+        msg_data.content,
+        template=f"{{}}{subtext}",
+        filter=_convert_nitro_emojis,
+    )
     if file:
         msg_data.attachments.append(file)
         content += "\n-# (content attached)"
@@ -65,12 +90,17 @@ async def move_message_via_webhook(
 
 
 def format_or_file(
-    message: str, *, template: str | None = None
+    message: str,
+    *,
+    template: str | None = None,
+    filter: Callable[[str], str] | None = None,
 ) -> tuple[str, discord.File | None]:
     if template is None:
         template = "{}"
+    if filter is None:
+        filter = lambda s: s
 
-    if len(full_message := template.format(message)) > 2000:
+    if len(full_message := filter(template.format(message))) > 2000:
         return template.format(""), discord.File(
             BytesIO(message.encode()), filename="content.md"
         )
--- a/app/utils/webhooks.py
+++ b/app/utils/webhooks.py
@@ -17,12 +17,16 @@ GuildTextChannel = discord.TextChannel | discord.Thread
 _EMOJI_REGEX = re.compile(r"<(a?):(\w+):(\d+)>", re.ASCII)
 
 
+def get_ghostty_guild() -> discord.Guild:
+    return next(g for g in bot.guilds if "ghostty" in g.name.casefold())
+
+
 def _convert_nitro_emojis(content: str, *, force: bool = False) -> str:
     """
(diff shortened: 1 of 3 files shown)


app/utils/webhooks.py
```diff
@@ -1,11 +1,38 @@
...
 
+from app.setup import bot
 from app.utils.message_data import MessageData, scrape_message_data
 
 GuildTextChannel = discord.TextChannel | discord.Thread
 
+_EMOJI_REGEX = re.compile(r"<(a?):(\w+):(\d+)>", re.ASCII)
+
+
+def _convert_nitro_emojis(s: str, *, force: bool = False) -> str:
```
00-kat, line 15: Not sure where to put it.  It's unlikely to be used outside of the `move_message_via_webhook()` function, but it also doesn't feel right leaving it in `webhook.py`.  I can't put it in `__init__.py` as that would cause a circular dependency.

app/utils/webhooks.py
```diff
@@ -1,11 +1,38 @@
...
+    """
+    Converts a custom emoji to a concealed hyperlink.  Set `force` to True
+    to convert emojis in the current guild too.
+    """
+    guild = next(g for g in bot.guilds if "ghostty" in g.name.casefold())
```
trag1c, line 20: Maybe this could be its own util too since it's the same as `entity_mentions/fmt.py:33`? `get_guild()` or whatever

app/utils/webhooks.py
trag1c: I think it's fine to keep it here. If you feel like making it more visible, you can make it public and export it in `__init__.py` 👍

trag1c: Also, could the parameter be renamed to something more meaningful, like `content`?

app/utils/webhooks.py
```diff
@@ -1,11 +1,38 @@
...
+        animated = bool(match.group(1))
+        id_ = int(match.group(3))
+        emoji = bot.get_emoji(id_)
+        if not force and not animated and emoji and emoji.guild_id == guild.id:
+            return match.group(0)
+
+        ext = "gif" if animated else "webp"
+        tag = "&animated=true" * animated
+        name = match.group(2)
+        return f"[{name}](https://cdn.discordapp.com/emojis/{id_}.{ext}?size=48{tag}&name={name})"
```
trag1c, line 32: A few nits here (as separate diffs w/ a completed suggestion at the end for reading convenience):
* `re.Match` is parametrizable (`str` or `bytes`)
```diff
-def r(match: re.Match) -> str:
+def r(match: re.Match[str]) -> str:
```
* I personally prefer unpacking `Match.groups` calls if I use all groups. Also, the type conversions aren't necessary (other than the `get_emoji` call), so I think it's okay to omit them.
```diff
-animated = bool(match.group(1))
-id_ = int(match.group(3))
-emoji = bot.get_emoji(id_)
-name = match.group(2)
+animated, name, id_ = match.groups()
+emoji = bot.get_emoji(int(id_))
```
* And when actually accessing specific groups, I prefer to use lookup syntax (it's shorter and apparently also nearly 2x as fast for some reason??)
```diff
-return match.group(0)
+return match[0]
```
* And maybe this sub could also use a better name:
```diff
-def r(match: re.Match[str]) -> str:
+def replace_nitro_emoji(match: re.Match[str]) -> str:
```
```suggestion
    def replace_nitro_emoji(match: re.Match[str]) -> str:
        animated, name, id_ = match.groups()
        emoji = bot.get_emoji(int(id_))
        if not force and not animated and emoji and emoji.guild_id == guild.id:
            return match[0]

        ext = "gif" if animated else "webp"
        tag = "&animated=true" * animated
        return f"[{name}](https://cdn.discordapp.com/emojis/{id_}.{ext}?size=48{tag}&name={name})"
```

app/utils/webhooks.py
```diff
@@ -1,11 +1,38 @@
+import re
+from collections.abc import Callable
```
trag1c, line 2: Could you throw this one under a `TYPE_CHECKING` block?

trag1c (changes requested):

app/utils/webhooks.py
00-kat: > for reading convenience

…and then GitHub proceeds to make the unified diff as bad as possible.  Thanks GitHub!!! /s

00-kat:
Not requesting another review as this is currently untested since I don't have Nitro (though it *should* work).

app/utils/webhooks.py
00-kat: > Also, the type conversions aren't necessary (other than the `get_emoji` call)

Actually, I multiply `animated` to a string later, so the `bool()` *was* necessary.  When I read the code I thought `animated` was an `int` not a `str` though, so I didn't catch that until I tested the code.

app/utils/webhooks.py
```diff
@@ -1,11 +1,42 @@
...
+
+import re
 from io import BytesIO
+from typing import TYPE_CHECKING
 
 import discord
 
+from app import utils
+from app.setup import bot
 from app.utils.message_data import MessageData, scrape_message_data
```
trag1c, line 11:
```suggestion
from app.setup import bot
from app.utils import get_ghostty_guild
from app.utils.message_data import MessageData, scrape_message_data
```

app/utils/webhooks.py
```diff
@@ -1,11 +1,42 @@
...
+    guild = utils.get_ghostty_guild()
+
+    def replace_nitro_emoji(match: re.Match[str]) -> str:
+        animated, name, id_ = match.groups()
+        emoji = bot.get_emoji(int(id_))
+        if not force and not animated and emoji and emoji.guild_id == guild.id:
+            return match[0]
+
+        ext = "gif" if animated else "webp"
+        tag = "&animated=true" if animated else ""
```
trag1c, line 35: Subjective, but:
```suggestion
        tag = animated and "&animated=true"
```

trag1c (changes requested):

app/utils/webhooks.py
00-kat: I could move the import inside the function if you want, along with a comment; I noticed that you(?) do that in some other parts of the code.

trag1c: Right, in that case I think the util can be moved to `webhooks.py`, it's getting re-exported properly either way.

PR #170, opened 2025-02-28
refactor(entity-mentions): use pydantic models for entities

trag1c:


app/components/entity_mentions/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+def state_validator(v: object) -> bool:
+    if isinstance(v, bool):
+        return v
+    if not isinstance(v, str):
+        msg = "`closed` must be a string or a bool"
+        raise TypeError(msg)
+    if v not in {"open", "closed"}:
+        msg = "`closed` must be 'open' or 'closed'"
+        raise ValueError(msg)
```
Uzaaft, line 15: Q: Couldnt this be a switch/match statement?

Uzaaft (approved):

app/components/entity_mentions/models.py
trag1c: It could! And it would look nicer as well, will refactor.

Issue #152, opened 2025-02-13
Ignore entity mentions in codeblocks


trag1c:
This might be finally feasible after #145 and its `zig_codeblocks.extract_codeblocks` get merged.

PR #123, opened 2025-01-08
feat: implement `/close` command

trag1c:
Closes #113.

Note that the following part wasn't implemented:
> (would require special logic for the case of a post being made by moving a message)

<sub>(I didn't feel like it + it made the entire feature kinda ugly + it just doesn't seem like it would happen often)</sub>

PR #172, opened 2025-02-28
feat: ignore entity mentions in codeblocks

mvares:
- Closes #171

### DIFF
--- a/app/components/entity_mentions/resolution.py
+++ b/app/components/entity_mentions/resolution.py
@@ -14,7 +14,7 @@ ENTITY_REGEX = re.compile(
     re.IGNORECASE,
 )
 
-CODEBLOCK_REGEX = re.compile(r'```.*?```', re.DOTALL)
+CODEBLOCK_REGEX = re.compile(r"```.*?```", re.DOTALL)
 
 
 class OwnerCache(TTRCache[str, str]):
@@ -24,8 +24,9 @@ class OwnerCache(TTRCache[str, str]):
 
 owner_cache = OwnerCache(3600)  # 1 hour
 
-def has_entity_mention(content):
-    text_outside_codeblocks = CODEBLOCK_REGEX.sub('', content)
+
+def has_entity_mention(content: str) -> bool:
+    text_outside_codeblocks = CODEBLOCK_REGEX.sub("", content)
 
     return ENTITY_REGEX.search(text_outside_codeblocks) is not None
 
--- a/app/components/entity_mentions/resolution.py
+++ b/app/components/entity_mentions/resolution.py
@@ -14,6 +14,8 @@ ENTITY_REGEX = re.compile(
     re.IGNORECASE,
 )
 
+CODEBLOCK_REGEX = re.compile(r'```.*?```', re.DOTALL)
+
 
 class OwnerCache(TTRCache[str, str]):
     async def fetch(self, key: str) -> None:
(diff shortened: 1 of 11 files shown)


00-kat:
You should probably rebase this on the new main by the way.

app/components/entity_mentions/resolution.py
```diff
@@ -14,6 +14,8 @@
     re.IGNORECASE,
 )
 
+CODEBLOCK_REGEX = re.compile(r"```.*?```", re.DOTALL)
```
00-kat, line 17: My regex is a little bit rusty, but is this not the same as `.*`?

```suggestion
CODEBLOCK_REGEX = re.compile(r"```.*```", re.DOTALL)
```

app/components/entity_mentions/resolution.py
```diff
@@ -23,6 +25,12 @@ async def fetch(self, key: str) -> None:
 owner_cache = OwnerCache(3600)  # 1 hour
 
 
+def has_entity_mention(content: str) -> bool:
+    text_outside_codeblocks = CODEBLOCK_REGEX.sub("", content)
```
00-kat, line 29: Note to future readers of this PR:
This simple check doesn't handle:
````
> ```
> #1234
```
````
That will show up with no codeblocks in Discord, but it will be excluded here.  I'm 90% certain that anyone who sends that likely forgot the last `> ` and *meant* to have it in a code block, so attempting to make this function *correct* will likely result in a worse outcome.

00-kat (approved):
I tried rebasing this on main and there didn't seem to be any conflicts, so whatever I read here is almost certainly going to be the same as what will exist after the commits here are cleaned up.

00-kat:
Oh also, based on your other commit messages, did you forget to add `chore` to the `run format` commit message?

app/components/entity_mentions/resolution.py
trag1c: This should probably be using `zig_codeblocks.extract_codeblocks` instead anyway.

00-kat:
By the way, this fails for the case where there are entity mentions both outside and inside the code block:
````markdown
#123```#456```
````

00-kat:
> Closes https://github.com/ghostty-org/discord-bot/pull/171

Wait hold on, what was wrong with #171?  Also you meant “Closes #152” right?

mvares:
> Wait hold on, what was wrong with #171? Also you meant “Closes #152” right?

Problems with git merge.

00-kat:
> Maybe our option would be:

I'm pretty sure the issue is here: https://github.com/ghostty-org/discord-bot/blob/c69c85fd26c7421f1ea9da34857702529bb4886f/app/components/entity_mentions/resolution.py#L31

That returns a boolean, which means if there is even one entity mention outside a codeblock, this code reacts to every single one (including those in codeblocks): https://github.com/ghostty-org/discord-bot/blob/c69c85fd26c7421f1ea9da34857702529bb4886f/app/core.py#L80

I believe the correct solution would be to change this: https://github.com/ghostty-org/discord-bot/blob/c69c85fd26c7421f1ea9da34857702529bb4886f/app/components/entity_mentions/resolution.py#L47
So that it only ever loops over the ones that are not in a codeblock.

00-kat (changes requested):
If you're not planning to work on this anymore, would you mind if I were to continue it in a new PR?

mvares:
Go ahead

00-kat:
Superseded by #187.

PR #173, opened 2025-03-01
Add an activity status

00-kat:
It only takes so long for a bike shed to be flooded with string (that is, I expect the exact string to be bikeshed).

Some other ideas:
- “Competing in **moderation with GearBot**”
- “Playing **Ghostty 👻**” (ghost emoji optional)

I tried setting the timestamp to when the bot started up, but for some reason Discord did not like it and it didn't show up anywhere.

Edit: quick note, if your suggestion starts with `Playing`, `Streaming`, `Listening to`, `Watching` or `Competing in`, it'll can be a special status, otherwise it'll just be the average custom status that shows up in a thought bubble on the profile.

### DIFF
--- a/app/core.py
+++ b/app/core.py
@@ -34,10 +34,19 @@ from app.utils import is_dm, is_mod, try_dm
 @bot.event
 async def on_ready() -> None:
     await load_emojis()
+
     if not autoclose_solved_posts.is_running():
         autoclose_solved_posts.start()
+
     bot_status.last_login_time = dt.datetime.now(tz=dt.UTC)
+
     print(f"Bot logged on as {bot.user}!")
+
+    await bot.change_presence(
+        activity=discord.Activity(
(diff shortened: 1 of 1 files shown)


matthew-hre:
Nothing I can think of isn't cringy, so I'll vote towards "Watching over the Ghostty server"

Some cringe-worthy ones, just to fuel the discussions:
Haunting your threads
Watching for posts in showcase
Listening for your issue numbers
Playing with my config file

rhodes-b:
Playing your friendly ghost (reverence to casper)
Playing pacman (homage to old logo)
(+1 the cfg file option above)

maybe worth adding a few statuses  mods can set with a command

00-kat:
Okay c'mon now GitHub.

00-kat:
Superseded by #179.

PR #175, opened 2025-03-02
Add Pyright

00-kat:
Not sure why it wasn't done before, so I'm just experimenting with it here.  There are *a lot* of lints that have to be fixed for this.  Also, mypy is *insanely slow* and *insanely resource-hungry* (on my machine: no cache: 1G mem, maxes out one thread, 1.5 minutes; with cache: 816M, maxes out one thread, 7 seconds).

### DIFF
--- a/app/components/docs.py
+++ b/app/components/docs.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime as dt
 import json
+from collections.abc import Iterable
 from typing import NotRequired, TypedDict, cast
 
 import discord
@@ -98,10 +99,13 @@ async def section_autocomplete(
 async def page_autocomplete(
     interaction: discord.Interaction, current: str
 ) -> list[Choice[str]]:
-    if not (interaction.data and (options := interaction.data.get("options"))):
+    if not interaction.data:
+        return []
+    options = cast(Iterable[dict[str, str]] | None, interaction.data.get("options"))
+    if not options:
         return []
     section = next(
-        (cast(str, opt["value"]) for opt in options if opt["name"] == "section"),
+        (opt["value"] for opt in options if opt["name"] == "section"),
         None,
     )
     if section is None:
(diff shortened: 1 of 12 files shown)


app/components/zig_codeblocks.py
```diff
@@ -209,8 +217,8 @@ async def zig_codeblock_edit_handler(
     if not (new_contents or new_files or before in frozen_messages):
         # All code was edited out
         codeblock_linker.unlink(before)
-        for reply in replies:
-            await reply.delete()
+        for r in replies:
```
00-kat, line 220: I hate this so much too.  Honestly no clue what to do about this one, this was the easiest solution I could conjure up.

app/components/docs.py
```diff
@@ -98,10 +99,13 @@ async def section_autocomplete(
 async def page_autocomplete(
     interaction: discord.Interaction, current: str
 ) -> list[Choice[str]]:
-    if not (interaction.data and (options := interaction.data.get("options"))):
+    if not interaction.data:  # TODO(Kat): needed?
```
00-kat, line 102: `interaction.data` is guaranteed to be a `dict`, and `{}` is the only `dict` that coerces to `False`, and the `data.get` below would handle this case just fine (since `{}.get(x)` returns `None` for all `x`).

00-kat:
With the exception of a singular `TODO`, this pull request should be ready now.  The question now is, *should it be merged*?

app/components/entity_mentions/cache.py
```diff
@@ -7,6 +7,16 @@
 from .models import Entity, Issue, PullRequest
 from app.setup import gh
 
+__all__ = [
+    "CacheKey",
+    "Entity",
```
00-kat, line 12: Maybe I should only export the ones mypy complains about instead of all of them?

app/components/docs.py
00-kat: As discovered on Discord: it's a `dict` at runtime and in the docs, but it's typed as an `Optional[InteractionData]`.  Thanks discord.py?  I'm just removing the check.

00-kat: > I'm just removing the check.

Yeah no, that won't work; mypy still sees it as what the type hint says.

00-kat:
Haha okay no 55s vs 7s I think we know who's winning here.

00-kat:
Okay c'mon GitHub what.  I did the rename from the web interface there's no way you still close it…

00-kat:
Superseded by #177 (branch rename).

PR #179, opened 2025-03-05
Add an activity status

00-kat:
Continuation of #173, thanks as usual GitHub 🙄.

It only takes so long for a bike shed to be flooded with string (that is, I expect the exact string to be bikeshed).

I tried setting the timestamp to when the bot started up, but for some reason Discord did not like it and it didn't show up anywhere.

Edit: quick note, if your suggestion starts with `Playing`, `Streaming`, `Listening to`, `Watching` or `Competing in`, it'll can be a special status, otherwise it'll just be the average custom status that shows up in a thought bubble on the profile.

### DIFF
--- a/app/components/activity_status.py
+++ b/app/components/activity_status.py
@@ -1,0 +1,21 @@
+import secrets
+
+from discord import Activity, ActivityType, CustomActivity
+from discord.ext import tasks
+
+from app.setup import bot
+
+STATUSES = (
+    Activity(type=ActivityType.watching, name="over the Ghostty server 👻"),
+    CustomActivity(name="Haunting your threads 🧵"),
+    Activity(type=ActivityType.watching, name="posts in #showcase"),
+    Activity(type=ActivityType.watching, name="over #help"),
+    Activity(type=ActivityType.listening, name="your complaints"),
+    Activity(type=ActivityType.playing, name="with my config file"),
+    Activity(type=ActivityType.competing, name="the terminal game"),
+)
+
+
+@tasks.loop(hours=2)
+async def randomize_activity_status() -> None:
+    await bot.change_presence(activity=secrets.choice(STATUSES))
--- a/app/components/__init__.py
+++ b/app/components/__init__.py
@@ -1,5 +1,6 @@
 from . import (
     accept_invite,
+    activity_status,
     autoclose,
     close_help_post,
     docs,
@@ -12,6 +13,7 @@ from . import (
 
 __all__ = (
     "accept_invite",
+    "activity_status",
     "autoclose",
     "close_help_post",
     "docs",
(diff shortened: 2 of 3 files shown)


app/components/activity_status.py
```diff
@@ -0,0 +1,21 @@
+import random
```
trag1c, line 1: `secrets` instead for a little spice? 🔥

app/components/activity_status.py
```diff
@@ -0,0 +1,21 @@
...
+
+statuses = [
+    Activity(type=ActivityType.watching, name="over the Ghostty server 👻"),
+    Activity(type=ActivityType.custom, name="Haunting your threads 🧵"),
+    Activity(type=ActivityType.watching, name="posts in #showcase"),
+    Activity(type=ActivityType.watching, name="over #help"),
+    Activity(type=ActivityType.listening, name="your complaints"),
+    Activity(type=ActivityType.playing, name="with my config file"),
+    Activity(type=ActivityType.competing, name="the terminal game"),
+]
```
trag1c, line 16:
```suggestion
STATUSES = (
    Activity(type=ActivityType.watching, name="over the Ghostty server 👻"),
    Activity(type=ActivityType.custom, name="Haunting your threads 🧵"),
    Activity(type=ActivityType.watching, name="posts in #showcase"),
    Activity(type=ActivityType.watching, name="over #help"),
    Activity(type=ActivityType.listening, name="your complaints"),
    Activity(type=ActivityType.playing, name="with my config file"),
    Activity(type=ActivityType.competing, name="the terminal game"),
)
```

app/components/activity_status.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@tasks.loop(hours=6)
```
trag1c, line 19: Why 6 hours? 🤔 I imagined this would be something like 30 minutes or at least 1 hour 😄

app/components/activity_status.py
00-kat: Seems like a waste of the system's randomness source to me, but sure.

app/components/activity_status.py
trag1c: Gotcha, how's 2 hours?

00-kat: [Fine by me](https://github.com/ghostty-org/discord-bot/compare/ffd57e3848ba1c008af9b080a3031c4b1d356756..5c70a03f2a1bf348e6936e875df0db790f5d7135).

trag1c (changes requested):
Could you include this module in `components/__init__.py`?

trag1c (commented):
Also, it looks like the custom `Haunting your threads 🧵` activity doesn't work. If it's picked the bot has no status at all.

00-kat:
I couldn't get the emoji to work for some reason (it got silently ignored), so I'm leaving it in the string itself.

trag1c (approved):

PR #180, opened 2025-03-06
Move replies and forwards

00-kat:


trag1c:
I'd say the cases of "new pin", "new thread", and "new followed server" are too unusual and niche to support. An error sounds better (or, for #155, a silent ignore).

00-kat:
> (or, for https://github.com/ghostty-org/discord-bot/issues/155, a silent ignore)

Oh I didn't even think about that actually 😅.  I've added a reminder to #155's description.

00-kat:
> are too unusual and niche to support

Less so than unusual or niche; I had thought it makes no sense to move them, since the pin/webhook/thread doesn't move with the message so it would end up being outright incorrect.

00-kat:
> so it would end up being outright incorrect

Completely by chance, I accidentally got something which was quite accurate, by changing this one line:
```diff
diff --git a/app/utils/message_data.py b/app/utils/message_data.py
index 3efa1c5368..96caf4083d 100644
--- a/app/utils/message_data.py
+++ b/app/utils/message_data.py
@@ -18,7 +18,7 @@
 
 async def scrape_message_data(message: discord.Message) -> MessageData:
     return MessageData(
-        message.content,
+        message.system_content,
         message.channel,
         *await _get_attachments(message),
         _get_reactions(message),
```
([Relevant docs].)

With a little bit of tweaking, I could make all of them actually reasonable; for example, here is the pin added event right now:
A screenshot.

However, looking through [the system message types][types], it's clear that most of these make zero sense to move.  I am hence going to disable moving of system messages in general, I think.  The current pin added one is also a little unintuitive (that's not what it looked like before), and is a bit of extra special-casing, so I don't think I'm going to support pin added either.

[Relevant docs]: https://discordpy.readthedocs.io/en/stable/api.html#discord.Message.system_content
[types]: https://discordpy.readthedocs.io/en/stable/api.html#discord.MessageType

app/components/move_message.py
```diff
@@ -71,7 +99,7 @@ def __init__(self, author: discord.Member, channel: GuildTextChannel) -> None:
     @discord.ui.button(
         label="Ghostping",
         emoji="👻",
-        style=discord.ButtonStyle.secondary,
+        style=discord.ButtonStyle.gray,
```
00-kat, line 102: Note: this was changed to be more consistent with `app.utils.DeleteMessage`.

00-kat:
>  Poll result: just as broken as before, except since it's also a reference you get an empty reply embed afterwards too!

There's no reason to not move these, but I honestly don't care enough to make them work ¯\\\_(ツ)\_/¯.

00-kat:
Rebased this on #186 since it had some changes which I had to account for but didn't want to add a new commit; that has to be merged first now.

app/utils/webhooks.py
```diff
@@ -104,7 +287,7 @@ def _format_subtext(
         else:
             lines.append(line)
     if skipped := msg_data.skipped_attachments:
-        lines.append(f"(skipped {skipped} large attachment(s))")
+        lines.append(f"(skipped {skipped} large attachment{'s' * (skipped != 1)})")
```
00-kat, line 290: (Yet another unrelated change—I personally despise programs which use `(s)` when determining the suffix is trivial.)

app/utils/webhooks.py
```diff
@@ -69,10 +127,163 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+            if message.type == discord.MessageType.context_menu_command
+            else "↪️ Reply"
+        )
+    )
+
+
+def _format_interaction(message: discord.Message) -> str:
+    if not message.interaction_metadata:
+        return message.content
+    with catch_warnings(action="ignore", category=DeprecationWarning):
```
00-kat, line 279: This just doesn't want to work for some reason; something has to be done about this before this is merged.

app/utils/webhooks.py
```diff
@@ -69,10 +127,166 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+        for attachment in msg_data.attachments
+        if Path(attachment.filename).suffix in SUPPORTED_IMAGE_FORMATS
+    ]
+    image_only_embeds = [
+        embed
+        for embed in embeds
+        if not embed.title
+        if not embed.description
+        if not embed.fields
+        if embed.image
```
trag1c, line 230:
```suggestion
        if embed.image and not (embed.title or embed.description or embed.fields)
```

app/utils/webhooks.py
```diff
@@ -69,10 +127,166 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+    if embeds or len(msg_data.attachments) > (1 if images else 0):
+        embed.add_field(
+            name="", value="-# (other forwarded content is attached)", inline=False
+        )
+
+    for line in _format_subtext(
+        None,
+        msg_data,
+        include_timestamp=False,
+    ).splitlines()[1:]:
```
trag1c, line 251: Why the `[1:]`?

app/utils/webhooks.py
```diff
@@ -69,10 +127,166 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+def _format_interaction(message: discord.Message) -> str:
+    if not message.interaction_metadata:
+        return message.content
+    # HACK: Message.interaction is deprecated, and discord.py disables any
+    # warning filter resulting in a bunch of warnings spammed in the logs even
+    # if it is ignored. There is no other way to get the name, and
+    # Message._interaction is not marked deprecated. Delectable.
+    if hasattr(message, "_interaction") and message._interaction is not None:  # pyright: ignore [reportPrivateUsage] # noqa: SLF001
+        prefix = "/" * (message.type != discord.MessageType.context_menu_command)
+        name = f"`{prefix}{message._interaction.name}`"  # pyright: ignore [reportPrivateUsage] # noqa: SLF001
```
trag1c, line 285: I'm pretty sure we don't need the `hasattr()` call?
```suggestion
    if message._interaction is not None:  # pyright: ignore [reportPrivateUsage] # noqa: SLF001
        prefix = "/" * (message.type != discord.MessageType.context_menu_command)
        name = f"`{prefix}{message._interaction.name}`"  # pyright: ignore [reportPrivateUsage] # noqa: SLF001
```
Also, since `MessageInteraction`s seem to always be truthy, I think we can bind it to avoid another warning suppress 2 lines later:
```suggestion
    if interaction := message._interaction:  # pyright: ignore [reportPrivateUsage] # noqa: SLF001
        prefix = "/" * (message.type != discord.MessageType.context_menu_command)
        name = f"`{prefix}{interaction.name}`"
```

app/utils/webhooks.py
```diff
@@ -130,11 +356,38 @@ async def move_message_via_webhook(
...
+    try:
+        ref = await _get_reference(message)
+    except discord.errors.NotFound:
+        embeds.append(_format_missing_reference(message))
+    else:
+        if ref is not None:
+            assert message.reference is not None
+            if message.reference.type == discord.MessageReferenceType.forward:
+                forward_embeds, forward_attachments = await _format_forward(ref)
+                embeds = forward_embeds + embeds
```
trag1c, line 381:
```suggestion
                embeds = [*forward_embeds, *embeds]
```

app/utils/webhooks.py
00-kat: > I'm pretty sure we don't need the `hasattr()` call?

It's a private field, and it could be removed in the future when `Message.interaction` is removed.

app/utils/webhooks.py
00-kat: Is there any reason I'm not aware of other than style?

app/utils/webhooks.py
00-kat: `_format_subtext` adds an extra newline at the start:
```python-console
>>> lines = ["foo", "bar", "baz"]
>>> "".join(f"\n-# {line}" for line in lines)
'\n-# foo\n-# bar\n-# baz'
>>> _.splitlines()
['', '-# foo', '-# bar', '-# baz']
```
I'll add a comment though.

app/utils/webhooks.py
trag1c: https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

app/utils/webhooks.py
```diff
@@ -69,10 +127,166 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+    description = reply.content
+    ref_exists = True
+    try:
+        ref = await _get_reference(reply)
+    except discord.errors.NotFound:
+        ref = discord.utils.MISSING
+        ref_exists = False
+    if ref is not None:
+        assert reply.reference is not None
+        if reply.reference.type == discord.MessageReferenceType.forward:
```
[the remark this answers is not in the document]
00-kat: Shall I change `discord.MessageType` comparisons too?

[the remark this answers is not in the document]
00-kat: I'm also not sure where you got so many line numbers from, `rg '== discord.Message'` yields five.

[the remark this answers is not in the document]
trag1c: A general rule for comparing enum members 😄 `is` is preferred, just like for other singletons (e.g. `None`).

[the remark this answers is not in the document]
00-kat: Yeah that's what I had guessed 👍️.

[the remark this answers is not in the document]
00-kat: Ah figured out where the other two were from (this time `rg ==`'d).

app/utils/webhooks.py
```diff
@@ -28,6 +33,62 @@ def get_ghostty_guild() -> discord.Guild:
...
+        if message.reference.type != discord.MessageReferenceType.forward:
+            # This check is subtly different from the one at the top. If we
+            # have a reference that's not a forward, we don't want to continue,
+            # of course. *But*, unlike up top where we return the reference, we
+            # want to return the current message itself, as otherwise we would
+            # be returning the reply the last forward is replying to rather than
+            # the last forward itself.
+            return message
+        message = ref
+        ref = await _get_original_message(message)
```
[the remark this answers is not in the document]
00-kat: Ah that one was from before I made it a walrus, fixed.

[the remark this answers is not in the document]
00-kat: Actually I lied, I added it in a different commit but don't remember doing so; no clue why I added it.

app/utils/webhooks.py
00-kat: Actually no, I'll change it to add the newline at the call site instead.

00-kat:
Looks like GitHub does not reopen a PR if the branch reappears.  Guess I'm going to  have to make a new one… oops.

00-kat:
Wait no I have a reopen button again.

00-kat:
…not anymore?

00-kat:
Okay so I *do* have a reopen button, but it is greyed out unless I reopen #189, and if I do it doesn't let me click it because #189 is open…

00-kat:
…and it doesn't do that anymore, just as I was recording a video.

![\[visible confusion\]](https://tenor.com/view/obi-wan-confusion-gif-10824788)

PR #186, opened 2025-03-22
Include the moved message's timestamp

00-kat:
Closes section 4 of #155.

A screenshot.

### DIFF
--- a/app/utils/webhooks.py
+++ b/app/utils/webhooks.py
@@ -81,7 +81,6 @@ def dynamic_timestamp(dt: dt.datetime, fmt: str | None = None) -> str:
 
 
 def _format_subtext(
-    message: discord.Message,
     executor: discord.Member | None,
     msg_data: MessageData,
     *,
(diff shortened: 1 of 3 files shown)


trag1c:
While sometimes a useful piece of information, I think this clutters the message too much (especially bad for short ones). I would consider it an improvement if this part only showed up when a significant (to be defined) amount of time has passed, e.g. seeing that the original message was 2 days old when moved is much more useful than let's say a 7-minute one.

00-kat:
> significant (to be defined) amount of time

How does twelve hours sound?

trag1c:
Fine by me :+1:

00-kat:
Dun.

00-kat:
I would prefer for this to be merged soon by the way, since #180 can't be merged until this is merged (and I just rebased it onto this too).

app/utils/webhooks.py
```diff
@@ -75,13 +75,35 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
     lines: list[str] = []
     if reactions := msg_data.reactions.items():
         lines.append("   ".join(f"{emoji} x{count}" for emoji, count in reactions))
+    if message.created_at > dt.datetime.now(tz=dt.UTC) - dt.timedelta(hours=12):
+        include_timestamp = False
+    if include_timestamp:
+        line = dynamic_timestamp(message.created_at)
+        if message.edited_at is not None:
+            line += f" (edited at {dynamic_timestamp(message.edited_at, 't')})"
+        lines.append(line)
```
trag1c, line 99: Let's make `created_at` and `edited_at` fields of `MessageData` so that we don't have to pass the `discord.Message` separately.

trag1c (changes requested):

trag1c (approved):

PR #187, opened 2025-03-23
Ignore entity mentions in code blocks

### LINKED ISSUE #152: Ignore entity mentions in codeblocks

00-kat:
Closes #152; supersedes #172.

See #172 for discussion and TODOs.


00-kat:
Ookay this was *much much* smaller than I expected, I'm in disbelief.

Please don't merge it until the TODO is fixed; I'll be fixing it as soon as @trag1c publishes zig-codeblocks 0.3.0.

app/components/entity_mentions/resolution.py
```diff
@@ -23,6 +25,14 @@ async def fetch(self, key: str) -> None:
 owner_cache = OwnerCache(3600)  # 1 hour
 
 
+def remove_codeblocks(content: str) -> str:
+    return reduce(
+        lambda acc, cb: acc.replace(f"```{cb.lang or ''}{cb.body}```", ""),
```
00-kat, line 30: A parameter of `acc` is pretty standard but it's probably confusing for someone who has never seen something like this before.  I can't think of any better names though.

I could also switch this back to a for loop I guess.

trag1c (approved):

PR #188, opened 2025-03-23
test: lay groundwork for test suite

trag1c:


README.md
```diff
@@ -134,11 +134,13 @@ This bot runs on Python 3.12+ and is managed with [uv]. To get started:
    ```console
    $ uv run -m app
    ```
-3. After you've made your changes, run the linter, formatter and type-checker:
+3. After you've made your changes, run the required checks:
    ```console
    $ uv run ruff check
    $ uv run ruff format
    $ uv run pyright app
+   $ uv run taplo fmt pyproject.toml
```
00-kat, line 142: Wow, a TOML formatter.

Random fun fact: TOML's creator was opposed to the idea of automatically formatting TOML, which is the reason why there does not exist any official formatter.

tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+import pytest
+
+from tests.fixtures import bot_env, safe_environ
+
+__all__ = ("bot_env",)
+
+
+def pytest_configure(config: pytest.Config) -> None:  # noqa: ARG001
```
00-kat, line 10: Can't you `_` it instead?
```suggestion
def pytest_configure(_: pytest.Config) -> None:
```
```suggestion
def pytest_configure(_config: pytest.Config) -> None:
```

tests/fixtures/config.py
```diff
@@ -0,0 +1,36 @@
+from types import SimpleNamespace
+
+import pytest
+
+BOT_ENV = SimpleNamespace(
+    BOT_TOKEN="fake-discord-token",
+    GITHUB_ORG="ghostty-org",
+    GITHUB_REPOS={"main": "ghostty", "bot": "discord-bot", "web": "website"},
+    GITHUB_TOKEN="fake-gh-token",
+    BOT_ACCEPT_INVITE_URL="https://discord.com/",
```
00-kat, line 10: Funnier URL.
```suggestion
    BOT_ACCEPT_INVITE_URL="https://youtu.be/-8KqaxE7kD0",
```

tests/test_close_help_post.py
```diff
@@ -0,0 +1,73 @@
...
+from githubkit.exception import RequestFailed
+
+from tests.utils import kitposer as kp
+
+from app.components.close_help_post import mention_entity
+from app.components.entity_mentions.discussions import DISCUSSION_QUERY
+from app.components.entity_mentions.models import GitHubUser
+
+
+def fill_entity(**kwargs: Any) -> SimpleNamespace:
```
00-kat, line 15: Wouldn't it be a better idea to have this in some utils file or something?  I can imagine that something similar to this would be needed for the other GitHub tests too.

tests/utils/kitposer.py
```diff
@@ -0,0 +1,76 @@
...
+        context = f"{self._context}/{attr}"
+        return type(self)(self.responses, context)
+
+    def __call__(self, *a: object, **kw: object) -> Any:
+        if self._context not in self.responses:
+            msg = f"Unexpected call for context {self._context!r}"
+            raise PoserSetupError(msg)
+
+        blackbox = self.responses[self._context]
+        # Hack: Fetching the original key instead of doing "Call(*a, **kw) in blackbox"
```
00-kat, line 60:
```suggestion
        # HACK: Fetching the original key instead of doing "Call(*a, **kw) in blackbox"
```

tests/test_close_help_post.py
```diff
@@ -0,0 +1,73 @@
...
+    defaults: dict[str, Any] = {
+        "number": 0,
+        "title": "",
+        "body": None,
+        "html_url": "https://github.com/foo/bar/issues/1",
```
00-kat, line 20: Self promo is very yum 😋.
```suggestion
        "html_url": "https://github.com/trag1c/zig-codeblocks/issues/1",
```
Also I didn't expect it but it coincidentally happens to be about testing too 💀.

tests/test_close_help_post.py
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return SimpleNamespace(**defaults | kwargs)
+
+
+# Calling __new__ to skip initialization and having to pass in a proper Reponse object
+failed_request = RequestFailed.__new__(RequestFailed)
+repo = ("ghostty-org", "ghostty")
+gh_env = kp.KitPoser(
+    {
+        "/rest/pulls/async_get": {
```
00-kat, line 37: I cannot figure out where these `/`s are parsed, but if it's in your code and not GitHubKit itself, it would be nice if this could be `.`s (or any other character) instead, to avoid confusion with GitHub API URLs (which are also `/rest`, etc.).  Unless these are actually GitHub API URLs, though I doubt it since HTTP doesn't have async functions.

00-kat (changes requested):
Mostly useless review comments, enjoy /s.

And of course, I have never used PyTest before, so this isn't really a proper review; for all i know `pytest.fixture` is a fancy way to fix your hard drive by driving away the French (`rm -fr /`).  None of the code *looks* incorrect to me though.

uv.lock
```diff
@@ -157,6 +157,15 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/38/fc/bce832fd4fd99766c04d1ee0eead6b0ec6486fb100ae5e74c1d91292b982/certifi-2025.1.31-py3-none-any.whl", hash = "sha256:ca78db4565a652026a4db2bcdf68f2fb589ea80d0be70e03929ed730746b84fe", size = 166393 },
 ]
 
+[[package]]
+name = "colorama"
```
00-kat, line 161: (Took us long enough 😝.)

tests/conftest.py
trag1c: Nope, `pytest` expects that the parameter is named `config`.

tests/test_close_help_post.py
trag1c: Very likely! I felt like waiting until another module needs it though.

tests/test_close_help_post.py
trag1c: The goal of KitPoser is to not use githubkit or make any HTTP requests at all, so these aren't actually used there.

These could be `.`s but I felt like slashed worked well since the logic is not complex enough to remove the leading one (so it'd have to be ".rest.pulls.async_get" which kinda looks weird to me).

tests/conftest.py
00-kat: Oh wow, okay.

tests/test_close_help_post.py
00-kat: It didn't look too weird to me, probably because of relative import, but fair.  I don't particularly mind as long as it's not `rest::pulls::async<get>` or something.

00-kat (approved):

PR #189, opened 2025-03-24
Move replies, forwards and interactions, and disallow moving system messages

00-kat:
Continuation of #180, because I accidentally deleted the branch momentarily, sorry @trag1c.

Closes sections 5 and 6 of #155.

See #180 for the (now finished) check list.

### DIFF
--- a/app/utils/webhooks.py
+++ b/app/utils/webhooks.py
@@ -28,6 +28,48 @@ def get_ghostty_guild() -> discord.Guild:
         raise ValueError(msg) from None
 
 
+async def _get_original_message(message: discord.Message) -> discord.Message | None:
+    """Can throw discord.errors.NotFound if the original message was deleted."""
+    if (msg_ref := message.reference) is None:
+        return None
+    if msg_ref.cached_message is not None:
+        return msg_ref.cached_message
+    if message.guild is None or msg_ref.message_id is None:
+        return None
+    channel = message.guild.get_channel(msg_ref.channel_id)
+    if not isinstance(channel, discord.TextChannel):
+        return None
+    return await channel.fetch_message(msg_ref.message_id)
+
+
+async def _get_reference(message: discord.Message) -> discord.Message | None:
+    ref = await _get_original_message(message)
+    if ref is None:
+        # There was no reference whatsoever.
+        return None
+    assert message.reference is not None
+    if message.reference.type != discord.MessageReferenceType.forward:
+        # We don't have a forward, fantastic, we're done here.
+        return ref
+    # And now, we have a forward. Discord doesn't collapse forwarded forwards,
+    # so we shall do it ourselves. This loop should not run for replies, as if
+    # that happens we would end up dereferencing all the way back to the start
+    # of a reply chain, which would be a horendous idea.
+    message = ref
+    while (ref := await _get_original_message(message)) is not None:
+        assert message.reference is not None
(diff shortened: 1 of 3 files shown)


00-kat:
Follow up from https://github.com/ghostty-org/discord-bot/pull/180#discussion_r2010031216, I forgot about `!=`; fixed now.

app/utils/webhooks.py
```diff
@@ -65,14 +134,167 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
-        text="Unable to attach sticker."
+
+def truncate(s: str, length: int, *, suffix: str = "…") -> str:
+    if len(s) <= length:
+        return s
+    return s[: length - len(suffix)] + suffix
+
+
+async def _format_reply(
+    reply: discord.Message, message_map: dict[int, str] | None = None
```
trag1c, line 151: `message_map` here (and in `_format_context_menu_command` and `_format_forward`) seems to never be non-None, is this dead code?

app/utils/webhooks.py
```diff
@@ -65,14 +134,167 @@ async def _get_sticker_embed(sticker: discord.StickerItem) -> discord.Embed:
...
+    description = reply.content
+    ref_exists = True
+    try:
+        ref = await _get_reference(reply)
+    except discord.errors.NotFound:
+        ref = discord.utils.MISSING
+        ref_exists = False
+    if ref is not None:
+        assert reply.reference is not None
+        if reply.reference.type is discord.MessageReferenceType.forward:
```
trag1c, line 166: Could you flip this condition?

app/utils/webhooks.py
00-kat: It's dead right now, but I'll need this later for bulk-moving messages; I'll remove it for now and put it on a separate branch though.

app/utils/webhooks.py
00-kat: That was a very intentional decision of mine: there are more than two message types that are considered a reference, and I made sure that every comparision to that enumeration used `forward` and not `reply` (I originally didn't pay much attention to this, but I did go back and fix them all (I think) up).

Pins, context menu commands, etc. are grouped into the same enumeration entry as replies, but I didn't want people to be confused about whether this code was about only replies, or about everything grouped into the same bucket as replies.  And while this is in `_format_reply`, I don't give *only* replies to this function; `_format_context_menu_command` below is just a wrapper around this one, after all.

I *could* change it to `if reply.reference.type is not discord.MessageReferenceType.default`, but I thought `forward` reads better than `default`.

00-kat: >  And while this is in `_format_reply`, I don't give *only* replies to this function; `_format_context_menu_command` below is just a wrapper around this one, after all.

Though I guess the parameter to this function being named `reply` doesn't help.

trag1c: Um 😅 I meant changing
```py
if reply.reference.type is discord.MessageReferenceType.forward:
    nine_lines = of_code("here")
```
to
```py
if reply.reference.type is not discord.MessageReferenceType.forward:
    continue
nine_lines = of_code("here")
```

trag1c: lol I thought this code was in a loop for some reason and wanted to suggest `if reply is not forward: continue`, my bad

trag1c (approved):
Great work!

PR #190, opened 2025-03-25
Reduce the number of regexes used

00-kat:
I'm sorry I just had to.


trag1c (approved):
Fine… 😄

This technically behaves differently for things with consecutive capital letters (for `MyHTTP` the regex produces `My HTTP` and this change produces `My H T T P`), but it doesn't matter here.

00-kat:
Oh yeah let me change that real quick.

PR #191, opened 2025-03-25
Make the SGR regex “more correct”, so to speak

00-kat:
Not like it matters, but I just had to.

### DIFF
--- a/app/components/zig_codeblocks.py
+++ b/app/components/zig_codeblocks.py
@@ -25,7 +25,7 @@ MAX_CONTENT = 51_200  # 50 KiB
 MAX_ZIG_FILE_SIZE = 8_388_608  # 8 MiB
 VIEW_TIMEOUT = 60.0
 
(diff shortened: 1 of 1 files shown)


trag1c:
The SGR sequences this is meant to operate on, as you may know, will never use `:`s. I'd prefer to keep the regex conservative if that's all it needs.

00-kat:
I was thinking if someone were to move this pattern to, say, `utils/` without knowing that they *may* find themselves with a weird bug.  Of course this is hypothetical as heck and anyway barely anyone uses `:`.  Close at your own discretion.

trag1c:
I'm closing this but I'll add a comment there to make its purpose clearer.

00-kat:
> barely anyone uses `:`

(For some stats, https://github.com/termstandard/colors?tab=readme-ov-file#terminal-emulators.)

Issue #183, opened 2025-03-14
Special-case XKCD mentions


00-kat:
People often request this, https://discord.com/channels/1005603569187160125/1320902256656388116/1346822180813410357 and https://discord.com/channels/1005603569187160125/1320881498848497755/1350043756644139039.  No repository would be shadowed by this change either.

PR #192, opened 2025-03-26
Small changes to TTRCache

00-kat:
1. Made the TTRCache splat its arguments to timedelta.
2. Moved it to `utils/` as I expect it would be useful outside of `entity_mentions` too (e.g., for #183).

I recommend viewing every commit individually, as it makes it more clear what changed than if you only look at the final diff.

### DIFF
--- a/app/utils/cache.py
+++ b/app/utils/cache.py
@@ -1,0 +1,35 @@
+import datetime as dt
+from abc import ABC, abstractmethod
+
+
+class TTRCache[KT, VT](ABC):
+    def __init__(self, **ttr: float) -> None:
+        """Keyword arguments are passed to datetime.timedelta."""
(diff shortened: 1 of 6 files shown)


app/utils/cache.py
```diff
@@ -0,0 +1,35 @@
+import datetime as dt
+from abc import ABC, abstractmethod
+
+
+class TTRCache[KT, VT](ABC):
+    def __init__(self, **ttr: float) -> None:
```
00-kat, line 6: This was previously `int` but is now `float`; the reason for this change is because timedelta accepts `float` and I see no reason why `float`s would not make sense in this class.

trag1c (approved):
I like the `**ttr`!

PR #193, opened 2025-03-26
Lock old posts to discourage necroposting

### LINKED ISSUE #174: Discourage necroposting
Make the bot give a heads up when people comment on an old help post. Kinda like what Refined GitHub does:

---

As suggested by @hqnna:
> check last timestamp when they post, if it's surpassed the necroposting threshold lock (not close) the post and notify the user.
> [...]
> smth like
> 
> hey @​user, this post hasn't had activity for \<threshold\>.
> please make a new thread instead of using this one, thanks!.
> <sub>in order to prevent further necroposting this thread will now be locked.</sub>

00-kat:
Closes #174.

A known issue is that sending multiple messages in quick succession will trigger multiple responses, but I do not know of any clean way to work around this as the Discord API is not the fastest thing in the world, and this operation is inherently racy.

Documentation update: #197.

### DIFF
--- a/app/components/lock_old_posts.py
+++ b/app/components/lock_old_posts.py
@@ -5,7 +5,7 @@ import datetime as dt
 import discord
 
 from app import config
-from app.utils import dynamic_timestamp, post_is_solved
+from app.utils import aenumerate, dynamic_timestamp, post_is_solved
 
 
 async def check_for_old_posts(message: discord.Message) -> None:
@@ -21,20 +21,19 @@ async def check_for_old_posts(message: discord.Message) -> None:
     ):
         return
 
-    try:
-        # Ignore messages less than a minute old for the same person may have
-        # sent multiple in quick succession, resulting in their own message
-        # being considered as activity.
-        last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
-    except IndexError:
-        # There were no messages older than one minute. Assuming the threshold
-        # below stays at a reasonably large limit, this either means that every
-        # message was deleted, or that there was no response and the original
-        # poster deleted the starter message. It's very unlikely for the post
-        # to no longer be relevant; nor is it likely that anyone would even
-        # message in such a thread. Thus, silently exit if this is the case.
-        return
-    if last.created_at > now - dt.timedelta(days=30):
+    # Ignore messages less than a minute old for the same person may have
+    # sent multiple in quick succession, resulting in their own message
+    # being considered as activity.
+    last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
+    # Don't lock the post if it isn't old enough.
+    if last is None or last.created_at > now - dt.timedelta(days=30):
+        # If last was None, there were no messages older than one minute.
+        # Assuming the threshold above stays at a reasonably large limit, this
+        # either means that every message was deleted, or that there was no
+        # response and the original poster deleted the starter message. It's
+        # very unlikely for the post to no longer be relevant; nor is it likely
+        # that anyone would even message in such a thread. Thus, silently exit
+        # if this is the case.
         return
 
     try:
@@ -57,8 +56,6 @@ async def _get_message(
     *,
     before: discord.abc.SnowflakeTime | None = None,
     around: discord.abc.SnowflakeTime | None = None,
-) -> discord.Message:
-    return [
-        message
-        async for message in thread.history(limit=n + 1, before=before, around=around)
-    ][n]
+) -> discord.Message | None:
+    messages = thread.history(limit=n + 1, before=before, around=around)
+    return await anext((m async for i, m in aenumerate(messages) if i == n), None)
--- a/app/components/lock_old_posts.py
+++ b/app/components/lock_old_posts.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import datetime as dt
+
+import discord
+
+from app import config
+from app.utils import dynamic_timestamp, post_is_solved
+
+
+async def check_for_old_posts(message: discord.Message) -> None:
+    post = message.channel
+    now = dt.datetime.now(tz=dt.UTC)
+    if (
+        not isinstance(post, discord.Thread)
+        or not post.parent
+        or post.parent.id != config.HELP_CHANNEL_ID
+        or post.locked
+        or post.last_message_id is None
+        or not post_is_solved(post)
+    ):
+        return
+
+    try:
+        # Ignore messages less than a minute old for the same person may have
+        # sent multiple in quick succession, resulting in their own message
+        # being considered as activity.
+        last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
+    except IndexError:
+        # There were no messages older than one minute. Assuming the threshold
+        # below stays at a reasonably large limit, this either means that every
+        # message was deleted, or that there was no response and the original
+        # poster deleted the starter message. It's very unlikely for the post
+        # to no longer be relevant; nor is it likely that anyone would even
+        # message in such a thread. Thus, silently exit if this is the case.
+        return
+    if last.created_at > now - dt.timedelta(days=30):
+        return
+
+    try:
+        creation_time_ago = dynamic_timestamp(
+            (await post.fetch_message(post.id)).created_at, "R"
+        )
+    except discord.NotFound:
+        creation_time_ago = "over a month ago"
+    await message.reply(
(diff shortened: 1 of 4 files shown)


app/components/lock_old_posts.py
```diff
@@ -0,0 +1,54 @@
...
+    await post.edit(locked=True)
+
+
+async def _get_message(
+    thread: discord.Thread,
+    n: int,
+    /,
+    *,
+    before: discord.abc.Snowflake | dt.datetime | None = None,
+    around: discord.abc.Snowflake | dt.datetime | None = None,
```
trag1c, line 45: We can use the `discord.abc.SnowflakeTime` type alias if we use `from __future__ import annotations`.
```suggestion
    before: discord.abc.SnowflakeTime | None = None,
    around: discord.abc.SnowflakeTime | None = None,
```

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,54 @@
...
+        "relevant. Please open a new thread instead, making sure to provide "
+        "the required information."
+    )
+    await post.edit(locked=True)
+
+
+async def _get_message(
+    thread: discord.Thread,
+    n: int,
+    /,
```
trag1c, line 42: I wouldn't lock those parameters to be positional-only.

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,54 @@
...
+import discord
+
+from app import config
+from app.utils import dynamic_timestamp, post_is_solved
+
+
+async def check_for_old_posts(message: discord.Message) -> None:
+    post = message.channel
+    now = dt.datetime.now(tz=dt.UTC)
+    two_fortnights_ago = now - dt.timedelta(weeks=4)
```
trag1c, line 12: Can we please make this something like this instead? 😵‍💫 
```suggestion
    stale_threshold = now - dt.timedelta(days=30)
```

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,54 @@
...
+    around: discord.abc.Snowflake | dt.datetime | None = None,
+    strict: bool = False,
+) -> discord.Message:
+    messages = [
+        message
+        async for message in thread.history(limit=n + 1, before=before, around=around)
+    ]
+    if messages or strict:
+        return messages[n]
+    return [message async for message in thread.history(limit=n + 1)][n]
```
trag1c, line 54: What purpose does this double `thread.history` fallback(?) serve?

trag1c (changes requested):

app/components/lock_old_posts.py
00-kat: Lol fine 😅.

app/components/lock_old_posts.py
00-kat: If there aren't enough messages before a given time then the indexing in the `return` would fail, so unless instructed otherwise this function falls back to trying without `before` and `around`.  I wasn't sure where to put it though, putting such a fallback at the call site seemed uglier to me.

00-kat: Funnily enough Copilot happened to comment on the exact thing this fallback was intended to solve.

00-kat: …and I just remembered that it's possible to delete the starter message so the invariant I had assumed doesn't actually exist.  Thanks Copilot I guess.

00-kat: I've moved the check to the call site now.

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,63 @@
...
+        last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
+    except IndexError:
+        # There were no messages older than one minute. Assuming the treshold
+        # below stays at a reasonably large limit, this either means that every
+        # message was deleted, or that there was no response and the original
+        # poster deleted the starter message. It's very unlikely for the post
+        # tono longer be relevant; nor is it likely that anyone would even
+        # message in such a thread. Thus, silently exit if this is the case.
+        return
+    if last.created_at > now - dt.timedelta(seconds=30):
```
trag1c, line 36: Is this possible?

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,63 @@
...
+        or not post_is_solved(post)
+    ):
+        return
+    try:
+        # Ignore messages less than a minute old for the same person may have
+        # sent multiple in quick succession, resulting in their own message
+        # being considered as activity.
+        last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
+    except IndexError:
+        # There were no messages older than one minute. Assuming the treshold
```
trag1c, line 29:
```suggestion
        # There were no messages older than one minute. Assuming the threshold
```

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,63 @@
...
+    thread: discord.Thread,
+    n: int,
+    *,
+    before: discord.abc.SnowflakeTime | None = None,
+    around: discord.abc.SnowflakeTime | None = None,
+) -> discord.Message:
+    return [
+        message
+        async for message in thread.history(limit=n + 1, before=before, around=around)
+    ][n]
```
trag1c, line 63: 👇 maybe? So that we don't have to store all messages if we only want the n-th one (even though the only usage is a 2-elem case).
```suggestion
    messages = (
        m async for m in thread.history(limit=n + 1, before=before, around=around)
    )
    return islice(messages, n, None)
```

app/components/lock_old_posts.py
00-kat: Kinda.  It's possible *in theory* but generally Bruce or I send a message in every help post so it's super unlikely.

I've seen quite a few posts with no messages and no info other than the title before though.

app/components/lock_old_posts.py
00-kat: Oh didn't know about `islice`.

app/components/lock_old_posts.py
```diff
@@ -0,0 +1,64 @@
...
+        last = await _get_message(post, 1, before=now - dt.timedelta(minutes=1))
+    except IndexError:
+        # There were no messages older than one minute. Assuming the threshold
+        # below stays at a reasonably large limit, this either means that every
+        # message was deleted, or that there was no response and the original
+        # poster deleted the starter message. It's very unlikely for the post
+        # to no longer be relevant; nor is it likely that anyone would even
+        # message in such a thread. Thus, silently exit if this is the case.
+        return
+    if last.created_at > now - dt.timedelta(seconds=30):
```
00-kat, line 37: Oops I forgot to change this back to days, leaving this here so I don't forget again.
```suggestion
    if last.created_at > now - dt.timedelta(days=30):
```

app/components/lock_old_posts.py
00-kat: I forgot to test it before I don't think `islice` works with `AsyncGenerator`?

app/components/lock_old_posts.py
00-kat: Oops I forgot to change this back to days.

app/components/lock_old_posts.py
trag1c: Yeah that's my bad :( There aren't too many built-in async iter utils, so our best bet would probably be to write something like this (since there's no `aenumerate`):
```py
i = 0
async for message in thread.history(...):
    if i == n:
        return message
    i += 1
raise RuntimeError # unreachable or something
```
Or just accept the reality of having a list there 😞

00-kat: > since there's no `aenumerate`

I guess I'll make one then.

app/utils/__init__.py
```diff
@@ -119,6 +121,24 @@ async def try_dm(account: Account, content: str, **extras: Any) -> None:
...
+
+def post_is_solved(post: discord.Thread) -> bool:
+    return any(
+        post_has_tag(post, tag)
+        for tag in ("solved", "moved to github", "duplicate", "stale")
+    )
+
+
+async def aenumerate[T](it: AsyncIterator[T]) -> AsyncIterator[tuple[int, T]]:
+    i = 0
```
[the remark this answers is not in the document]
00-kat: Oh I didn't even know `enumerate` had `start`.

At some point after writing `aenumerate` I was looking at it and thinking “maybe I should add a `start` param or something, that might be useful” then decided not to do it because I had believed `enumerate` didn't have one…

trag1c (approved):

PR #195, opened 2025-03-29
Use a webhook for /docs

### LINKED ISSUE #194: Use webhooks for `/docs`

00-kat:
Closes #194.

Also handles messages that are too long.  This affects both the interaction[^1] and the webhook paths, but I only noticed it now and don't want to make a new PR; plus you can't get 2000 characters without specifying `message`.

[^1]: It is still possible to get the interaction response with `message` specified in some specific channel types (such as voice channels).

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,9 @@ message option:
   <img src="https://github.com/user-attachments/assets/0938881f-80ad-44d0-8414-915324f2761e" alt="/docs command message option" height="250px">
 </p>
 
+If a message is provided, a webhook will be used to send the message under the
+interactor's server profile.
+
 ## `/close`
 
 A command group to mark help channel posts as resolved, with various options for different resolution scenarios:
--- a/app/components/docs.py
+++ b/app/components/docs.py
@@ -140,6 +140,10 @@ async def docs(
         await interaction.response.send_message("Documentation linked.", ephemeral=True)
     except ValueError as exc:
         await interaction.response.send_message(str(exc), ephemeral=True)
+    except discord.HTTPException:
+        await interaction.response.send_message(
+            "Message content too long.", ephemeral=True
+        )
 
 
 def get_docs_link(section: str, page: str) -> str:
--- a/app/components/docs.py
+++ b/app/components/docs.py
@@ -9,6 +9,7 @@ from discord.app_commands import Choice, autocomplete
 
 from app.components.status import bot_status
 from app.setup import bot, config, gh
+from app.utils import get_or_create_webhook
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -178,6 +178,9 @@ message option:
   <img src="https://github.com/user-attachments/assets/0938881f-80ad-44d0-8414-915324f2761e" alt="/docs command message option" height="250px">
```
00-kat, line 178: Probably needs a new image now.

app/components/docs.py
```diff
@@ -125,11 +126,28 @@ async def docs(
...
+        ):
+            webhook = await get_or_create_webhook(
+                "Ghostty Moderator", interaction.channel
+            )
+            await webhook.send(
+                f"{message}\n{get_docs_link(section, page)}",
+                username=interaction.user.display_name,
+                avatar_url=interaction.user.display_avatar.url,
+            )
+            await interaction.response.send_message(
```
00-kat, line 140: If it's possible to respond to an interaction without explicitly providing a message, that would be really nice too.

README.md
```diff
@@ -178,6 +178,9 @@ message option:
...
 </p>
 
+If a message is provided, a webhook will be used to send the message under the
+interactor's username.
```
trag1c, line 182: I think this should be clearer in that it's also using their avatar + technically it's their display name, not their username.

app/components/docs.py
```diff
@@ -125,11 +126,28 @@ async def docs(
     interaction: discord.Interaction, section: str, page: str, message: str = ""
 ) -> None:
     try:
-        await interaction.response.send_message(
-            f"{message}\n{get_docs_link(section, page)}"
-        )
+        if message and isinstance(
+            interaction.channel, discord.TextChannel | discord.ForumChannel
+        ):
```
trag1c, line 131: Why is this `isinstance` check now here? Also I'd much rather have the condition flipped since the `else` block is just one line.

00-kat: > Why is this `isinstance` check

As far as I can tell you can't create a webhook on other channel types, and Pyright complains.

> now here?

What do you mean?  Is this check also present somewhere else?

> Also I'd much rather have the condition flipped since the `else` block is just one line.

True, I'm surprised I didn't notice that.

README.md
00-kat: I can't think of a good way to say that, does `interactor's server profile` sound decent enough?

trag1c: Sounds good if it's actually their server profile and not their global profile.

app/components/docs.py
trag1c: > What do you mean? Is this check also present somewhere else?

No I just meant that it wasn't present in the code before, but yeah I figured it's for webhook reasons.

README.md
00-kat: I'm using `display_name` and `display_avatar.url`, so *in theory* yeah it is…  Not sure how much I trust Discord about this though.

README.md
trag1c: I'll commit one when this is merged.

app/components/docs.py
```diff
@@ -125,11 +126,24 @@ async def docs(
     interaction: discord.Interaction, section: str, page: str, message: str = ""
 ) -> None:
     try:
-        await interaction.response.send_message(
-            f"{message}\n{get_docs_link(section, page)}"
+        if not message or not isinstance(
+            interaction.channel, discord.TextChannel | discord.ForumChannel
+        ):
```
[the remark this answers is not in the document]
00-kat: Yeah that formatting made me instantly undo it, me personally I'd pass.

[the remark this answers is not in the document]
trag1c: oki :(

README.md
00-kat: You could commit to my branch I think; I have the check-box enabled.

trag1c (approved):

PR #196, opened 2025-03-30
Name linked entities again

00-kat:
Recreation of #44.  I can't figure out when this was removed.

### DIFF
--- a/app/components/entity_mentions/resolution.py
+++ b/app/components/entity_mentions/resolution.py
@@ -8,9 +8,11 @@ from .cache import TTRCache
 from app.setup import config, gh
 
 ENTITY_REGEX = re.compile(
+    r"(?P<site>\bhttps?://github\.com/)?"
     r"(?P<owner>\b[a-z0-9\-]+/)?"
     r"(?P<repo>\b[a-z0-9\-\._]+)?"
-    r"#(?P<number>\d{1,6})(?!\.\d)\b",
+    r"(?P<sep>/(?:issues|pull|discussions)/|#)"
+    r"(?P<number>\d{1,6})(?!\.\d)\b",
     re.IGNORECASE,
 )
 
@@ -37,9 +39,17 @@ async def find_repo_owner(name: str) -> str:
 async def resolve_repo_signatures(content: str) -> AsyncIterator[tuple[str, str, int]]:
     valid_signatures = 0
     for match in ENTITY_REGEX.finditer(content):
+        site, sep = match["site"], match["sep"]
+        # Ensure that the correct separator is used.
+        if bool(site) == (sep == "#"):
+            continue
+        # NOTE: this *must* be after the previous check, as the number can be
+        # an empty string if an incorrect separator was used, which would
+        # result in a ValueError in the call to int().
         owner, repo, number = match["owner"], match["repo"], int(match["number"])
+
         match owner, repo:
-            case None, None if number < 10:
+            case None, None if number < 10 and not site:
                 # Ignore single-digit mentions like #1, (likely a false positive)
                 continue
             case None, None:
--- a/app/components/entity_mentions/resolution.py
+++ b/app/components/entity_mentions/resolution.py
@@ -2,6 +2,7 @@ import re
 from collections.abc import AsyncIterator
 from contextlib import suppress
 
+import discord
 from githubkit.exception import RequestFailed
 
 from .cache import TTRCache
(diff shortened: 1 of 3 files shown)


trag1c:
> I can't figure out when this was removed.

Probably around the release, since the reason for those was that no embed showed up when it was private 😄

00-kat:
Found it https://github.com/ghostty-org/discord-bot/commit/b6f4b94be576f056fecbf6f55c806fe49b0eb4ea.

00-kat:
It appears to have been removed in #74 with no rationale, though its *so close* to the release that your explanation above is the most plausible.

app/components/entity_mentions/resolution.py
```diff
@@ -34,10 +37,25 @@ async def find_repo_owner(name: str) -> str:
     )
 
 
-async def resolve_repo_signatures(content: str) -> AsyncIterator[tuple[str, str, int]]:
+async def resolve_repo_signatures(
+    message: discord.Message,
```
00-kat, line 41: Also: I changed this to take the entire message as, while this is technically not the best place to put it, trying to forward a boolean back to the caller was excessively annoying and required making the code at the call site much harder to read.

app/components/entity_mentions/resolution.py
```diff
@@ -2,15 +2,18 @@
...
 from githubkit.exception import RequestFailed
 
 from .cache import TTRCache
 from app.setup import config, gh
 
 ENTITY_REGEX = re.compile(
+    r"(?P<site>\bhttps?://github\.com/)?"
     r"(?P<owner>\b[a-z0-9\-]+/)?"
     r"(?P<repo>\b[a-z0-9\-\._]+)?"
-    r"#(?P<number>\d{1,6})(?!\.\d)\b",
```
00-kat, line 13: It isn't obvious in the diff, but `1,6` was changed to `,6`; this check has been moved to Python code below.

app/components/entity_mentions/resolution.py
```diff
@@ -2,15 +2,18 @@
...
+    r"(?P<sep>/(?:issues|pull|discussions)/|#)"
+    r"(?P<number>\d{,6})(?!\.\d)\b",
```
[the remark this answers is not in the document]
00-kat: Oh I entirely missed the check inside the match-case somehow, and also didn't think through this too much thinking “it's currently matching two numbers so I just need to subtract one right...?”…

app/components/entity_mentions/resolution.py
```diff
@@ -34,14 +37,26 @@ async def find_repo_owner(name: str) -> str:
...
+        site, sep = match["site"], match["sep"]
+        # Ensure that the correct separator is used.
+        if (site and sep == "#") or (not site and sep != "#"):
+            continue
+        # NOTE: this *must* be after the previous check, as the number can be
+        # an empty string if an incorrect separator was used, which would
+        # result in a ValueError in the call to int().
         owner, repo, number = match["owner"], match["repo"], int(match["number"])
+        # Ignore single-digit mentions like #1 (likely a false positive).
+        if not (site or owner or repo) and number < 10:
```
00-kat, line 54: You'll probably dislike that I moved the check outside, but I *really* do not want to get `site` into the match-case as it's just going to be `_` in almost every case.

trag1c: 😅 We don't need to pass `site` as the match target:
```py
case None, None if number < 10 and not site:
    # Ignore single-digit mentions like https://github.com/ghostty-org/discord-bot/issues/1, (likely a false positive)
    continue
```

00-kat: …oh.

00-kat: You saw *nothing* :3.

I should probably not write code when I'm tired as heck.

app/components/entity_mentions/resolution.py
```diff
@@ -34,12 +37,24 @@ async def find_repo_owner(name: str) -> str:
...
-async def resolve_repo_signatures(content: str) -> AsyncIterator[tuple[str, str, int]]:
+async def resolve_repo_signatures(
+    message: discord.Message,
+) -> AsyncIterator[tuple[str, str, int]]:
     valid_signatures = 0
-    for match in ENTITY_REGEX.finditer(content):
+    for match in ENTITY_REGEX.finditer(message.content):
+        site, sep = match["site"], match["sep"]
+        # Ensure that the correct separator is used.
+        if (site and sep == "#") or (not site and sep != "#"):
```
trag1c, line 47: Since this is basically an NXOR:
```suggestion
        if bool(site) == (sep == "#"):
```

trag1c (approved):

PR #199, opened 2025-04-03
Correct the supported Python versions in the README

00-kat:
Discord.py doesn't work on Python 3.13.


00-kat:
Never mind it seems like it *does* work now.

PR #202, opened 2025-04-08
Handle Nitro reactions in moved messages

00-kat:
As well as a few other assorted half-related commits >.>; I didn't want to make an entirely new PR for them.

Here's the image for commit two:
A screenshot of multiple moved messages containing masked links to Nitro emoji URLs that have enormous embeds attached to them.

### DIFF
--- a/app/utils/message_data.py
+++ b/app/utils/message_data.py
@@ -18,7 +18,7 @@ class MessageData(NamedTuple):
     edited_at: dt.datetime | None
     attachments: list[discord.File]
     skipped_attachments: int
-    raw_reactions: list[discord.Reaction]
+    reactions: list[discord.Reaction]
 
 
 async def scrape_message_data(message: discord.Message) -> MessageData:
--- a/app/utils/message_data.py
+++ b/app/utils/message_data.py
@@ -19,7 +19,6 @@ class MessageData(NamedTuple):
     attachments: list[discord.File]
     skipped_attachments: int
     raw_reactions: list[discord.Reaction]
-    reactions: dict[str | discord.Emoji, int]
 
 
 async def scrape_message_data(message: discord.Message) -> MessageData:
@@ -30,7 +29,6 @@ async def scrape_message_data(message: discord.Message) -> MessageData:
         message.edited_at,
         *await _get_attachments(message),
         message.reactions,
-        _get_reactions(message),
     )
 
 
@@ -49,14 +47,3 @@ async def _get_attachments(message: discord.Message) -> tuple[list[discord.File]
         attachments.append(discord.File(fp, filename=attachment.filename))
 
     return attachments, skipped_attachments
-
-
-def _get_reactions(message: discord.Message) -> dict[str | discord.Emoji, int]:
-    reactions: dict[str | discord.Emoji, int] = {}
-    for reaction in message.reactions:
-        if isinstance(emoji := reaction.emoji, discord.Emoji) and not emoji.is_usable():
-            continue
-        if isinstance(emoji, discord.PartialEmoji):
-            continue
-        reactions[emoji] = reaction.count
-    return reactions
--- a/app/utils/message_data.py
+++ b/app/utils/message_data.py
@@ -18,6 +18,7 @@ class MessageData(NamedTuple):
     edited_at: dt.datetime | None
     attachments: list[discord.File]
     skipped_attachments: int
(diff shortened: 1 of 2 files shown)


app/utils/webhooks.py
```diff
@@ -297,8 +297,15 @@ def _format_subtext(
     include_timestamp: bool = True,
 ) -> str:
     lines: list[str] = []
-    if reactions := msg_data.reactions.items():
-        lines.append("   ".join(f"{emoji} x{count}" for emoji, count in reactions))
+    if reactions := msg_data.raw_reactions:
+        formatted_reactions = [
+            f"{emoji} x{reaction.count}"
+            if isinstance(emoji := reaction.emoji, str)
+            or (isinstance(emoji, discord.Emoji) and emoji.is_usable())
```
00-kat, line 304: And I also wish this could be a `hasattr` but Pyright doesn't like those.

app/utils/webhooks.py
```diff
@@ -297,8 +297,15 @@ def _format_subtext(
...
+            else f"[{emoji.name}](<{emoji.url}>) x{reaction.count}"
```
00-kat, line 305: Off-topic: should we consider Unicode multiplication signs instead?  [I mentioned this on Discord before](https://discord.com/channels/1281624934636064811/1281624935558807678/1340852949559279616) but I don't think you responded.

Some candidates:
- `×` — MULTIPLICATION SIGN
- `⨯` — VECTOR OR CROSS PRODUCT
- `✕` — MULTIPLICATION X

A random data point: the GTK Ghostty app runtime was using the vector/cross product sign for the cross sign in the resize overlay until a month or two ago.

trag1c: > I don't think you responded.

My bad!

> Here's how they look when not monospaced, in order: ×12 ⨯12 ✕12. The current `x` is: x12.

The last one doesn't play too well on Discord:

I'm fine with either of the first two though! ^^

app/utils/webhooks.py
00-kat: LOL that's certainly a solution…

app/utils/webhooks.py
00-kat: What glyphs were those in your screenshot, in order?

This is what it looks like to me, with what I assume to be the default font:
A screenshot of the aforementioned glyphs in my Discord.
> ×12 ⨯12 ✕12.

I do agree about the last one though, it's really quite large or bold depending on where I paste it.

trag1c: > What glyphs were those in your screenshot, in order?

Same order as your original comment.

00-kat: Odd.  Perhaps it's a case of Discord's font not having this glyph and the fallback chosen for me (probably Noto Sans or DejaVu) having a very different-looking glyph than the one chosen for you (probably some default macOS font).  Guess I'll go with `×` (MULTIPLICATION SIGN) then, as it's nicest among both our screenshots.

app/utils/webhooks.py
```diff
@@ -297,8 +297,15 @@ def _format_subtext(
     include_timestamp: bool = True,
 ) -> str:
     lines: list[str] = []
-    if reactions := msg_data.reactions.items():
-        lines.append("   ".join(f"{emoji} x{count}" for emoji, count in reactions))
+    if reactions := msg_data.raw_reactions:
+        formatted_reactions = [
+            f"{emoji} x{reaction.count}"
+            if isinstance(emoji := reaction.emoji, str)
```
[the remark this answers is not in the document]
trag1c: I don't mind it at all!

app/utils/message_data.py
```diff
@@ -18,6 +18,7 @@ class MessageData(NamedTuple):
     edited_at: dt.datetime | None
     attachments: list[discord.File]
     skipped_attachments: int
+    raw_reactions: list[discord.Reaction]
     reactions: dict[str | discord.Emoji, int]
```
trag1c, line 22: Looks like with your change the original `reactions` field is never used now?

00-kat: I know it's not used in `webhooks.py` but I was too lazy to comb through every user of `message_data` to see if it uses `reactions`, but… probably.

trag1c: I guess it should be removed then 😄

trag1c (approved):

PR #204, opened 2025-04-09
Don't die on very early status requests

### LINKED ISSUE #203: Prevent status report failure when triggered before first scan
Repro (I think):
1. Launch the bot
2. DM it `status` when it has connected but before `on_ready` is run.
3. It should crash, because when trying to generate the report it tries to generate a dynamic timestamp from a `None` value

00-kat:
Closes #203.

### DIFF
--- a/app/components/status.py
+++ b/app/components/status.py
@@ -34,6 +34,12 @@ class BotStatus:
     last_scan_results: tuple[dt.datetime, int, int] | None = None
     last_sitemap_refresh: dt.datetime | None = None
 
+    @property
+    def initialized(self) -> bool:
+        return all(
+            (self.last_login_time, self.last_sitemap_refresh, self.last_scan_results)
+        )
+
     def _get_scan_data(self) -> SimpleNamespace:
         # Avoid circular import
         from app.components.autoclose import autoclose_solved_posts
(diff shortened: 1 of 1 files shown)


app/components/status.py
```diff
@@ -94,6 +106,8 @@ def _get_commit_hash() -> str:
 
 
 def status_message() -> str:
+    if not bot_status.initialized:
+        return "The bot has not finished initializing yet; try again shortly."
```
00-kat, line 110: Perhaps this should be a full stop; I can't decide.

trag1c: I think it's fine as is.

app/components/status.py
```diff
@@ -34,6 +34,14 @@ class BotStatus:
     last_scan_results: tuple[dt.datetime, int, int] | None = None
     last_sitemap_refresh: dt.datetime | None = None
 
+    @property
+    def initialized(self) -> bool:
+        return not (
+            self.last_login_time is None
+            or self.last_sitemap_refresh is None
+            or self.last_scan_results is None
+        )
```
trag1c, line 43:
```suggestion
        return all(
            (self.last_login_time, self.last_sitemap_refresh, self.last_scan_results)
        )
```

trag1c (approved):

PR #205, opened 2025-04-09
Store the lock threshold in a variable

00-kat:
Some very DRY timestamp crackers… and puns so bad people are crackl- I mean, cackling…

Motivated by the second computation of this timestamp after the introduction of https://github.com/ghostty-org/discord-bot/commit/1da7a6dcce57def33f3d38beb7d469b9b8421d7b.

I've also snuck a comment about https://github.com/ghostty-org/discord-bot/commit/1da7a6dcce57def33f3d38beb7d469b9b8421d7b in there as well.

### DIFF
--- a/app/components/lock_old_posts.py
+++ b/app/components/lock_old_posts.py
@@ -11,7 +11,9 @@ from app.utils import aenumerate, dynamic_timestamp, post_is_solved
 async def check_for_old_posts(message: discord.Message) -> None:
     post = message.channel
     now = dt.datetime.now(tz=dt.UTC)
+    one_minute_ago = now - dt.timedelta(minutes=1)
     one_month_ago = now - dt.timedelta(days=30)
+
     if (
         not isinstance(post, discord.Thread)
         or not post.parent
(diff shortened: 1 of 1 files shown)


app/components/lock_old_posts.py
```diff
@@ -20,17 +23,21 @@ async def check_for_old_posts(message: discord.Message) -> None:
         or not post_is_solved(post)
         or (
             (starter := post.starter_message) is not None
-            and starter.created_at > now - dt.timedelta(days=30)
+            # Do a quick check for the post starter message, whose timestamp is
+            # readily available without fetching messages from Discord. This
+            # works around ratelimits due to a high volume of fetches when
+            # messages are sent in the #help channel.
```
trag1c, line 29: I'm not sure I like the wording—the comment seems to imply (to me) that this is _only_ a workaround, while in reality it's an early guard that helps narrow the relevant messages down and was just overlooked.

00-kat: True, `works around` for something that isn't just a work around isn't particularly good wording.  Does this seem okay?
```suggestion
            # Also check the post starter message, whose timestamp is
            # readily available without fetching messages from Discord. This
            # reduces ratelimits due to a high volume of fetches when
            # messages are sent in the #help channel.
```

trag1c: Yeah that looks okay

trag1c (approved):

PR #207, opened 2025-04-10
Add hyperlinks to all relevant parts of the code links message

00-kat:
Heard you like printf-style string interpolation once, so here's an alternate diff :P /s:
<details>
<summary>Click or tap to display spoilered content.</summary>

```diff
diff --git a/app/components/entity_mentions/code_links.py b/app/components/entity_mentions/code_links.py
index 0d17340727..0db7ac3873 100644
--- a/app/components/entity_mentions/code_links.py
+++ b/app/components/entity_mentions/code_links.py
@@ -89,16 +89,42 @@
 
 
 def _format_snippet(snippet: Snippet) -> str:
-    range_info = (
-        f"lines {snippet.range.start + 1}–{snippet.range.stop}"  # noqa: RUF001
-        if snippet.range.stop > snippet.range.start + 1
-        else f"line {snippet.range.start + 1}"
-    )
-    unquoted_path = urllib.parse.unquote(snippet.path)
     return (
-        f"`{unquoted_path}`, {range_info}"
-        f"\n-# Repo: `{snippet.repo}`, revision: `{snippet.rev}`"
-        f"\n```{snippet.lang}\n{snippet.body}\n```"
+        "[`%s`](<https://github.com/%s/blob/%s/%s>), %s"
+        "\n-# Repo: [`%s`](<https://github.com/%s>),"
+        " revision: [`%s`](<https://github.com/%s/tree/%s>)\n```%s\n%s\n```"
+        % (
+            urllib.parse.unquote(snippet.path),
+            snippet.repo,
+            snippet.rev,
+            snippet.path,
+            "[lines %s–%s](<https://github.com/%s/blob/%s/%s#L%s-L%s>)"  # noqa: RUF001
+            % (
+                snippet.range.start + 1,
+                snippet.range.stop,
+                snippet.repo,
+                snippet.rev,
+                snippet.path,
+                snippet.range.start + 1,
+                snippet.range.stop,
+            )
+            if snippet.range.stop > snippet.range.start + 1
+            else "[line %s](<https://github.com/%s/blob/%s/%s#L%s>)"
+            % (
+                snippet.range.start + 1,
+                snippet.repo,
+                snippet.rev,
+                snippet.path,
+                snippet.range.start + 1,
+            ),
+            snippet.repo,
+            snippet.repo,
+            snippet.rev,
+            snippet.repo,
+            snippet.rev,
+            snippet.lang,
+            snippet.body,
+        )
     )
```

</details>

PR #208, opened 2025-04-10
Differentiate between branches and revisions

00-kat:
Based on #207, merge that first.

This heuristic prefers false negatives over false positives.  Making an entire request to GitHub just for this one small thing seems wasteful, so there will definitely be cases where branches are considered revisions.  Maybe we have this information already; I'm not too familiar with GitHub's API.

### DIFF
--- a/app/components/github_integration/code_links.py
+++ b/app/components/github_integration/code_links.py
@@ -1,4 +1,5 @@
 import re
+import string
 import urllib.parse
 from collections.abc import AsyncIterator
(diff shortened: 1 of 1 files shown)


trag1c:
> so there will definitely be cases where branches are considered revisions

And also tags will be considered branches :-)

00-kat:
Oh yeah, tags exist…

trag1c (approved):

PR #210, opened 2025-04-12
Handle www.github.com

00-kat:
GitHub just redirects it back to github.com, but these links *are* theoretically possible, similar to the http:// case.

Reject at one's own discretion.  This probably shouldn't be merged first since it would probably conflict with other, more important PRs.

### DIFF
--- a/app/components/entity_mentions/code_links.py
+++ b/app/components/entity_mentions/code_links.py
@@ -18,7 +18,7 @@ from app.utils import (
 )
 
 CODE_LINK_PATTERN = re.compile(
-    r"https?://github\.com/([^/]+)/([^/]+)/blob/([^/]+)/([^\?#]+)(?:[^\#]*)?"
+    r"https?://(?:www\.)?github\.com/([^/]+)/([^/]+)/blob/([^/]+)/([^\?#]+)(?:[^\#]*)?"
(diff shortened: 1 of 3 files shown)


app/components/entity_mentions/code_links.py
```diff
@@ -18,7 +18,7 @@
 )
 
 CODE_LINK_PATTERN = re.compile(
-    r"https?://github\.com/([^/]+)/([^/]+)/blob/([^/]+)/([^\?#]+)(?:[^\#]*)?"
+    r"https?://(?:www\.)github\.com/([^/]+)/([^/]+)/blob/([^/]+)/([^\?#]+)(?:[^\#]*)?"
```
trag1c, line 21: You didn't make it optional, so any non-`www.` link doesn't match now.

00-kat: Oops forgot the `?` when rebasing 😅.

trag1c: You missed it in all three places ^^

00-kat: …huh.  So I had the `?` in *only* this one before I rebased, according to the diff.

trag1c (approved):

PR #211, opened 2025-04-13
components: rename entity_mentions to github_integration

trag1c:
This also makes the original entity mention feature a subpackage.


app/components/github_integration/comments/fetching.py
```diff
@@ -9,9 +9,11 @@
 from githubkit.exception import RequestFailed
 from zig_codeblocks import extract_codeblocks
 
-from app.components.entity_mentions.cache import entity_cache
-from app.components.entity_mentions.discussions import get_discussion_comment
-from app.components.entity_mentions.models import Comment, EntityGist, GitHubUser
+from app.components.github_integration.mentions.cache import entity_cache
+from app.components.github_integration.mentions.discussions import (
+    get_discussion_comment,
```
00-kat, line 14: Should these two go outside (i.e.: `app.components.github_integration.{cache,discussions}`)?

00-kat (approved):

app/components/github_integration/comments/fetching.py
trag1c: `cache.py`—not really, but yeah I think `discussions.py` should be split up (and part of it should land in `comments/discussions.py`). I'll do that in a separate PR later.

PR #218, opened 2025-04-30
Prefer `typing.Self`

00-kat:
Hopefully it works with Python 3.12 too, I'm too lazy to compile another copy.

### DIFF
--- a/app/components/docs.py
+++ b/app/components/docs.py
@@ -1,8 +1,6 @@
-from __future__ import annotations
-
 import datetime as dt
 import json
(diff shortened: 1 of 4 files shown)


trag1c (approved):
mypy seems to complain about TypedDict + Self, but strict pyright is fine with it 🤷

00-kat:
Interesting.  Luckily Mypy is the least of our concerns because it complains about a billion other things as well (mostly implicit re-exports), so it's not like anyone's going to enjoy using it (in theory…).

PR #219, opened 2025-05-01
Allow deleting moved messages

00-kat:
Closes half of section three of #155, and lays the groundwork for the other half.

It's not perfect, but I doubt the few edge cases would ever occur.


00-kat:
I'll deal with the CI failure tomorrow (I didn't catch it before because Ruff 0.9.10, which is what I use, handles it correctly, while Ruff 0.9.6, which is what the CI uses as it is what is in the lockfile, doesn't).

I would like a suggestion on what to do:
  - Bump Ruff.
  - Add a `noqa`.

app/utils/webhooks.py
```diff
@@ -306,17 +324,20 @@ def _format_subtext(
...
+        line = f"Authored by {msg_data.author.mention}"
+        if include_timestamp:
+            line += " on "
+        lines.append(line)
     if include_timestamp:
         line = dynamic_timestamp(msg_data.created_at)
         if msg_data.edited_at is not None:
             line += f" (edited at {dynamic_timestamp(msg_data.edited_at, 't')})"
-        lines.append(line)
+        lines[-1] += line
```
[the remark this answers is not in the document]
00-kat: got dam

I had not caught it because currently no caller disables `include_timestamp` but not also `include_author` 🤦‍♀️.  I should probably write tests for this function later…

.env.example
```diff
@@ -1,4 +1,5 @@
 BOT_ACCEPT_INVITE_URL=https://github.com/ghostty-org
+BOT_WEBHOOK_NAME="Ghostty Moderator"
```
00-kat, line 2: I can see a few other alternatives:
1. Shove this in `config.py` instead of `.env`.
2. Hardcode it in `get_or_create_webhook()`.

00-kat:
I half-heartedly apologise for touching so many random files for minor unrelated changes 😅.

tests/utils/kitposer.py
```diff
@@ -57,7 +57,7 @@ def __call__(self, *a: object, **kw: object) -> Any:
             raise PoserSetupError(msg)
 
         blackbox = self.responses[self._context]
-        # HACK: Fetching the original key instead of doing "Call(*a, **kw) in blackbox"
+        # HACK: fetching the original key instead of doing "Call(*a, **kw) in blackbox"
```
00-kat, line 60: The majority of these comments[^1] have a lower-case letter after the colon, but every single comment apart from the one in `tests` was written by me; I evidently have not been consistent with this either.

Your comment uses a capital letter.

Should I change all of them to use a capital letter, or should I do what I've done here: make them all use a lowercase letter?

[^1]: Where I define “these comments” as lines in `app` and `tests` that match the Rust `regex` crate's interpretation of the regex `# [A-Z]*(\(.*\))?:`, with all capture groups ignored (more precisely, with the entire regex wrapped in `(?:)`[^2]).
[^2]: This is done by `ripgrep`; if your regex is incorrect the regex printed in the error message contains this exactly as I have mentioned it here.

00-kat:
Note: there is currently a known issue, where moved moved messages cannot be deleted, but I don't want to fix it now since it's an exceptionally rare case and it requires parsing the subtext, which I started in another commit and am not in the mood to do the commit surgery to reorder into this one.

.env.example
trag1c: I think I like the second alternative the most

00-kat: Do you mean that in a “change it to the second approach” way, or a “if this were to be changed, prefer the second approach” way?

trag1c: The former ^^

app/utils/webhooks.py
```diff
@@ -491,3 +513,84 @@ def format_or_file(
...
+
+    try:
+        return await webhook.fetch_message(message.id, thread=thread)
+    except discord.Forbidden:
+        return None
+    except discord.NotFound:
+        return discord.utils.MISSING
+
+
+def _find_snowflake(content: str, type_: str) -> tuple[int | None, int | None]:
```
trag1c, line 555:
```suggestion
def _find_snowflake(content: str, type_: str) -> tuple[int, int] | tuple[None, None]:
```

app/utils/webhooks.py
```diff
@@ -491,3 +513,84 @@ def format_or_file(
...
+    else:
+        channel = message.channel
+        thread = discord.utils.MISSING
+
+    for webhook in await channel.webhooks():
+        if webhook.id == message.webhook_id:
+            break
+    else:
+        return discord.utils.MISSING
+    if webhook.name != "Ghostty Moderator":
```
trag1c, line 543: I don't think this should be hardcoded since changing it is now allowed.

app/utils/webhooks.py
```diff
@@ -491,3 +513,84 @@ def format_or_file(
...
+    if message.webhook_id is None or isinstance(
+        message.channel,
+        # These types can't even have a webhook.
+        discord.DMChannel | discord.GroupChannel | discord.PartialMessageable,
+    ):
+        return discord.utils.MISSING
+
+    if isinstance(message.channel, discord.Thread):
+        thread = message.channel
+        if (channel := message.channel.parent) is None:
```
trag1c, line 532:
```suggestion
        if (channel := thread.parent) is None:
```

app/utils/webhooks.py
```diff
@@ -491,3 +513,84 @@ def format_or_file(
...
+    # don't contain anything except the subtext in their `Message.content`.
+
+    # If we have a channel mention, the executor is present; discard that part
+    # so that the executor is not accidentally picked as the author.
+    _, pos = _find_snowflake(subtext, "#")
+    if pos is not None:
+        subtext = subtext[:pos]
+
+    snowflake, _ = _find_snowflake(subtext, "@")
+    return int(snowflake) if snowflake else None
```
trag1c, line 596:
```suggestion
    return snowflake
```

app/utils/webhooks.py
```diff
@@ -361,8 +372,13 @@ def format_skipped(skipped: int) -> str:
         return f"Skipped {skipped} large attachment{'s' * (skipped != 1)}"
 
     def format(self) -> str:
-        context = (
+        original_message_info = (
+            self.author,
+            " on " * bool(self.author and self.timestamp),
```
trag1c, line 377: I think using `bool()` doesn't make it much shorter than an if expression, so maybe let's use that this time?
```suggestion
            " on " if self.author and self.timestamp else "",
```

trag1c (changes requested):

app/utils/webhooks.py
00-kat: That is the case pretty much all the time though :P.

app/utils/webhooks.py
```diff
@@ -491,3 +509,86 @@ def format_or_file(
...
     return full_message, None
+
+
+async def get_moved_message(
+    message: discord.Message, *, webhook_name: str = "Ghostty Moderator"
+) -> discord.WebhookMessage | None:
+    """
+    Returns None if it could not be acquired, and discord.utils.MISSING if the
+    provided message is not a moved message.
+    """
```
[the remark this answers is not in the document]
00-kat: Yeah I definitely didn't like it either but I honestly just don't know how to do sum types nicely in Python…

I'll see if I can figure something out, or maybe I'll do `discord.WebhookMessage | MovedMessageLookupFailed` (where `MovedMessageLookupFailed` is an enumeration) or something.

app/utils/webhooks.py
```diff
@@ -304,8 +306,16 @@ def dynamic_timestamp(dt: dt.datetime, fmt: str | None = None) -> str:
 
 
 class _SubText:
+    # WARNING: get_moved_message_author_id() makes a lot of assumptions about
+    # the structure of the subtext; be careful when editing this as
+    # invalidating the previous expected structure can break editing and
+    # deletion of messages moved before the change was merged or ALLOW THE
+    # WRONG PERSON TO EDIT OR DELETE A MOVED MESSAGE, *even if* that function
+    # is updated to match the new structure, as this subtext change won't
+    # retroactively apply to previously moved messages on Discord's servers.
```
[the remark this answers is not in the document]
00-kat: Oh, I just realised, one problem with timestamp checks is that the timestamp would need to be updated when the bot is *redeployed* rather than when the change is made, which would be annoying to do correctly…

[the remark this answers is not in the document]
trag1c: > which would be annoying to do correctly

I think it would be best if I pushed a change to update the timestamp right before deploying (maybe I could move it a minute or two forward to account for the redeploy time). Worst case scenario there would be a small X-minute window of wrong edits/deletions being allowed, but users don't even know when a redeploy happens I believe.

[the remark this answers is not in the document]
00-kat: The hard part is remembering 😛.

trag1c (approved):

PR #228, opened 2025-05-04
Fix typos

00-kat:


tests/utils/kitposer.py
```diff
@@ -57,7 +57,7 @@ def __call__(self, *a: object, **kw: object) -> Any:
             raise PoserSetupError(msg)
 
         blackbox = self.responses[self._context]
-        # HACK: Fetching the original key instead of doing "Call(*a, **kw) in blackbox"
+        # HACK: fetching the original key instead of doing "Call(*a, **kw) in blackbox"
```
00-kat, line 60: The majority of these comments[^1][^3] have a lower-case letter after the colon, but every single comment apart from this one was written by me; I evidently have not been consistent with this either.

This comment uses a capital letter.

Should I change all of them to use a capital letter, or should I retain what I've currently done: make them all use a lowercase letter?

[^1]: Where I define “these comments” as lines in `app` and `tests` that match the Rust `regex` crate's interpretation of the regex `# [A-Z]*(\(.*\))?:`, with all capture groups ignored (more precisely, with the entire regex wrapped in `(?:)`[^2]).
[^2]: This is done by ripgrep; if your regex is incorrect the regex printed in the error message contains this exactly as I have mentioned it here.
[^3]: There are a few more introduced in #219.

trag1c: I don't think I care, do whatever you like more ^^

00-kat: Lower-case it is then 😅.

app/core.py
```diff
@@ -97,7 +97,7 @@ async def on_message(message: discord.Message) -> None:
         reply_with_code(message),  # Look for GitHub code links and reply with contents
         reply_with_comments(message),  # Check for entity comments and reply with embeds
         check_for_old_posts(message),  # Check for bumps of old help posts and lock them
-        handle_xkcd_mentions(message),  # Reply to xkcd mentions with the URL
+        handle_xkcd_mentions(message),  # Reply to XKCD mentions with the URL
```
trag1c, line 100: Actually, why uppercase?

00-kat: …I don't know, actually.  I thought XKCD was normally full caps, so every place except the two places here uses it as full caps.  I guess I'll change every other XKCD to xkcd then?

trag1c: [Preferably ^^](https://xkcd.com/about/)

00-kat: Oh.

I was literally just on that page but didn't see that.

Well time to force push again then 😅.

trag1c (approved):

app/core.py
00-kat: I feel the need to point this out (there is no way this wasn't intentional…):

PR #242, opened 2025-06-15
add a safeguard for accidental VS16 additions

trag1c:
Follow-up to #240.

### DIFF
--- a/tests/test_meta.py
+++ b/tests/test_meta.py
@@ -2,14 +2,14 @@ from pathlib import Path
 
 import pytest
 
-VS16 = b"\xef\xb8\x8f"
-VS16_SUPPRESS_COMMENT = b"# test: allow-vs16"
+VS16 = "\ufe0f"
+VS16_SUPPRESS_COMMENT = "# test: allow-vs16"
 
 
 def check_dir_for_vs16(path: Path) -> None:
     error_locations: list[tuple[Path, int]] = []
     for file in path.rglob("*.py"):
-        for line_no, line in enumerate(file.read_bytes().splitlines(), 1):
+        for line_no, line in enumerate(file.read_text().splitlines(), 1):
             if VS16 in line and VS16_SUPPRESS_COMMENT not in line:
                 error_locations.append((file, line_no))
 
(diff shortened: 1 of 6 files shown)


tests/test_meta.py
```diff
@@ -0,0 +1,39 @@
...
+def check_dir_for_vs16(path: Path) -> None:
+    error_locations: list[tuple[Path, int]] = []
+    for file in path.rglob("*.py"):
+        for line_no, line in enumerate(file.read_bytes().splitlines(), 1):
+            if VS16 in line and VS16_SUPPRESS_COMMENT not in line:
+                error_locations.append((file, line_no))
+
+    if error_locations:
+        raise AssertionError(
+            f"unsuppressed VS16 found in {len(error_locations)} lines:\n"
```
00-kat, line 18: This message should definitely point the receiver to one of:
- #240.
- https://github.com/ghostty-org/discord-bot/commit/798365ee17f10f483b112d51d0fce63308c66da5.
- Some comment that can be added to this file (I can write one up if you want), and this message asking the receiver to view this file.
- Some markdown file checked into the repository elsewhere, and this message asking the receiver to view that markdown file.

00-kat (changes requested):
Apologies for taking so long 😅 I got distracted a lot today.

It would be nice if `# test: allow-vs16` wasn't required, and it is possible to programmatically determine which which emojis need their VS removed using Unicode's emoji data.

However, Python's standard library's `unicodedata` module doesn't include emoji data; it only appears to include [these data files].

Which would mean we need to either do a web request every time the test is run (or maybe caching for seven days), or we need to check in the current version and update it when newer emoji standards are published.  I've also sent some information (and sample code) [over here on Discord] so please do check that too ^^.

Assuming neither is desirable, this is likely the best way forward.  Maybe a accompanying script that goes through and automatically removes the offending VSs would be useful too, though; and since that script likely won't be run much (unlike this test) and won't be run in CI, a web request is definitely fine there.

[these data files]: https://github.com/python/cpython/blob/60181f4ed0e48ff35dc296da6b51473bfc553d16/Tools/unicode/makeunicodedata.py#L48-L58
[over here on Discord]: https://discord.com/channels/1281624934636064811/1294988140645453834/1383962379125260409

tests/test_meta.py
trag1c: Good call! I think I very slightly prefer a module comment/docstring over a Markdown file, but I would be fine with that too.

> (I can write one up if you want)

Please, you're far better at this than me :smile:

trag1c:
> It may also be useful for other cases (such as reactions in #245).

How so? :)

trag1c:
Funny how a rebase just flagged a new change on `main` 😄

00-kat:
> Funny how a rebase just flagged a new change on `main` 😄

That one can be ignored too (it's the one that resulted in me finding #240).

tests/test_meta.py
```diff
@@ -0,0 +1,39 @@
...
+            + "\n".join(f"- {file}:{line_no}" for file, line_no in error_locations)
+        )
+
+
+def test_src_unwanted_vs16() -> None:
+    check_dir_for_vs16(Path(__file__).parents[1] / "app")
```
00-kat, line 24: This should probably check the tests too, since any test that compares emojis will probably get confused by variation selectors.
```suggestion
    src_dir = Path(__file__).parents[1]
    check_dir_for_vs16(src_dir / "app")
    check_dir_for_vs16(src_dir / "tests")
```

PR #244, opened 2025-06-18
bump dev deps, use new ruff formatting style

trag1c:


.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Lint code
         run: uv run ruff check --output-format github
       - name: Check Python formatting
-        run: uv run ruff format --diff
+        run: uv run ruff format --preview --diff
```
00-kat, line 28: Shouldn't you update the command in the README too?

00-kat (changes requested):

PR #245, opened 2025-06-19
Handle deletion of messages moved multiple times

00-kat:
Related: https://github.com/ghostty-org/discord-bot/pull/219#issuecomment-2850907522 and https://github.com/ghostty-org/discord-bot/pull/219#issuecomment-2986561948.

I didn't bother with `MOVED_MESSAGE_MODIFICATION_CUTOFF` here since any issues with parsing the subtext aren't going to actually cause many issues; if this fails, nothing notable breaks, with the exception of the extremely rare case when someone tries to delete a message that was moved multiple times.  It definitely doesn't let others delete messages they did not author, since the author in the subtext if this fails would be the webhook ID.


app/utils/webhooks.py
```diff
@@ -452,9 +452,24 @@ async def move_message_via_webhook(  # noqa: PLR0913
...
+        and include_move_marks
+        and isinstance(
+            (webhook_message := await get_moved_message(message)),
+            discord.WebhookMessage,
+        )
+        # Check if the message has an author, to reduce false positives.
+        and get_moved_message_author_id(webhook_message) is not None
+    ):
+        # Append the new move mark to the existing subtext.
+        message.content, _, subtext = message.content.rpartition("\n")
```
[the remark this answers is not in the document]
00-kat: This is actually subtly incorrect; if there is a newline, the entire content is actually the subtext, as it likely means this moved message didn't have content (e.g.: attachment or sticker only, forwards, etc.).  Sure there are cases where it isn't the subtext, but those are the same ones as with `get_moved_message()` and `get_moved_message_author_id()`.

app/utils/webhooks.py
```diff
@@ -452,9 +452,20 @@ async def move_message_via_webhook(  # noqa: PLR0913
...
-    s = _SubText(msg_data, executor, poll if message.poll is not None else None)
-    subtext = s.format() if include_move_marks else s.format_simple()
+    if executor is not None and include_move_marks and await is_moved_message(message):
+        # Append the new move mark to the existing subtext. With the exception
+        # of incredibly unlikely false positives that pass through
+        # get_moved_message() and get_moved_message_author_id(), when there is
+        # no newline in the content, the entire message content is actually the
+        # subtext, since the subtext is always present while content might not
+        # be (e.g. attachments only, sticker only, forward).
+        message.content, _, subtext = message.content.rpartition("\n")
```
trag1c, line 462: If I'm understanding correctly, this only grabs the last line of the subtext, right? I'd say it should be clarified, if so.

trag1c (commented):
Looks good, just one note again :)

00-kat:
I want to mention now, before I forget, that the current changes I'm doing don't handle the case where a Zig code block is sent, then reactions are added to it, then it is replaced with a webhook message, then reactions are added to the replaced message, then that replaced message is moved.  I think it's fairly obvious why I don't care about that case ;D.

tests/test_message_moving.py
```diff
@@ -1,15 +1,77 @@
...
+)
+def test_format_partial_emoji(*, name: str, animated: bool, id_: int) -> None:
+    url = f"https://cdn.discordapp.com/emojis/{id_}.{'gif' if animated else 'png'}"
+    assert (
+        _format_emoji(discord.PartialEmoji(name=name, animated=animated, id=id_))
+        == f"[{name}](<{url}>)"
+    )
+
+
+class FakeEmoji(discord.Emoji):
```
00-kat, line 54: There's probably some better way with Pytest, but I know nothing about Pytest.

app/utils/webhooks.py
```diff
@@ -384,6 +392,53 @@ def _sub_join(*strs: str) -> str:
         return "\n".join(f"-# {s}" for s in strs if s)
 
 
+class SplitSubText:
```
00-kat, line 395: I can't think of a better name, do you have any ideas?  The only other reasonable one I could think of was `ExistingSubText`.  But none of the ones I can think of hint toward `message.content` also being stored in it.

app/utils/webhooks.py
```diff
@@ -452,17 +507,34 @@ async def move_message_via_webhook(  # noqa: PLR0913
     else:
         poll = message.poll
 
-    # The if expression is to skip the poll ended message if there was no poll.
-    s = _SubText(msg_data, executor, poll if message.poll is not None else None)
-    subtext = s.format() if include_move_marks else s.format_simple()
+    if include_move_marks and await is_moved_message(message):
+        # Append the new move mark to the existing subtext. We cast as
+        # is_moved_message() guarantees that we have a moved message.
+        split_subtext = SplitSubText(cast("discord.WebhookMessage", message))
```
00-kat, line 513: I really do not like this cast and would prefer to subclass `WebhookMessage` as `MovedMessage`[^1] or something instead.  I'll wait until you give me your opinion on that though.

[^1]: Or maybe `MovedMessageWithAuthor`, or some other name if you decide you like subclassic and have a better suggestion.

app/utils/webhooks.py
00-kat, line 395: Oh and: I know an entire class may be overkill for this,  but I went with this instead of a function because I'll need `__init__()` and `format()` in the future for removing the subtext when editing messages, and I don't want to make this one function just to make it a class momentarily after.

app/utils/webhooks.py
00-kat: https://github.com/ghostty-org/discord-bot/blob/62970c7794f3673b370981df4e732859eaca4d48/app/utils/webhooks.py#L457

`format_simple()` doesn't include the author.

Currently only replaced Zig codeblocks.

tests/test_message_moving.py
00-kat: Is there any reason for using `Mock(return_value=...)` instead of `lambda: ...` (`lambda _: ...` for the `__str__` case)?

app/utils/webhooks.py
```diff
@@ -452,17 +507,35 @@ async def move_message_via_webhook(  # noqa: PLR0913
...
+            if "•" in content:
+                # Ensure the move mark stays at the end, so that appending to
+                # the move mark later in SplitSubText.update() doesn't make the
+                # result incorrect.
+                subtext, _, move_mark = content.rpartition(" • ")
+                content = f"{subtext} • Content attached • {move_mark}"
+            else:
+                content += " • Content attached"
+        else:
+            content = "-# Content attached"
```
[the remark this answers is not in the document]
00-kat: I knew it was possible but I couldn't figure out how 😅.

tests/test_message_moving.py
trag1c: I think it's more idiomatic and allows for some more "advanced" things, like call tracking.

app/utils/webhooks.py
```diff
@@ -384,6 +390,53 @@ def _sub_join(*strs: str) -> str:
...
+    @staticmethod
+    def _get_reactions(reaction_line: str) -> dict[str, int]:
+        if not reaction_line.startswith("-# "):
+            return {}
+        d: dict[str, int] = {}
+        for s in reaction_line.removeprefix("-# ").split("   "):
+            try:
+                emoji, count = _REACTION_REGEX.findall(s)[0]
+            except IndexError:
+                return {}
```
trag1c, line 415: Why do we suddenly return an empty dict if the matches suddenly stop?

00-kat: If any of them don't match, we probably don't have an actual reactions line.  I'll see if I can word a comment to describe this.

app/utils/webhooks.py
```diff
@@ -384,6 +390,56 @@ def _sub_join(*strs: str) -> str:
...
+            return {}
+        d: dict[str, int] = {}
+        for s in reaction_line.removeprefix("-# ").split("   "):
+            try:
+                emoji, count = _REACTION_REGEX.findall(s)[0]
+            except IndexError:
+                # If any of the reactions don't match, we don't have an actual
+                # reaction line; return an empty dictionary to ignore that line
+                # as it may be a similar line in the content itself.
+                return {}
```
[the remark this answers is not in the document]
00-kat: I stg I find a new method under `re` that does something subtly different which none of the other ones do every other time I try to use regexes in Python…

app/utils/webhooks.py
```diff
@@ -384,6 +390,56 @@ def _sub_join(*strs: str) -> str:
...
+            self.reactions.setdefault(emoji, 0)
+            self.reactions[emoji] += reaction.count
+
+    def format(self) -> str:
+        if not self.reactions:
+            return self.subtext
+        formatted_reactions = "   ".join([
+            f"{emoji} ×{count}"  # noqa: RUF001
+            for emoji, count in self.reactions.items()
+        ])
```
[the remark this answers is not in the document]
00-kat: Done in https://github.com/ghostty-org/discord-bot/compare/0f1d2613710c1e1808b265b2af5aa7e2531cff58..2cf5be8c11b057ef6356797406699484b8697496 (I also updated the one in `_format_reactions()` to match).

trag1c (approved):

PR #246, opened 2025-06-20
Avoid referring to the context menu as a “right click” menu

00-kat:
It's not right click on mobile, it's long press. And “context menu” reads more fluently than “right click or long press”.

### DIFF
--- a/app/components/message_filter.py
+++ b/app/components/message_filter.py
@@ -16,8 +16,8 @@ MESSAGE_DELETION_TEMPLATE = (
 )
 MESSAGE_CONTENT_NOTICE = "Here's the message you tried to send:"
 COPY_TEXT_HINT = (
-    "-# **Hint:** You can get your original message with formatting preserved "
-    'by using the "Copy Text" action on right-click.'
+    "-# **Hint:** you can get your original message with formatting preserved "
+    'by using the "Copy Text" action in the context menu.'
 )
 REGULAR_MESSAGE_TYPES = frozenset({
     discord.MessageType.default,
--- a/app/components/zig_codeblocks.py
+++ b/app/components/zig_codeblocks.py
@@ -158,7 +158,7 @@ async def _prepare_reply(
         )
 
     codeblocks = list(extract_codeblocks(message.content))
-    file_highlight_note = 'Click "View whole file" to see the highlighting.'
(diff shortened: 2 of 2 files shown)


app/components/message_filter.py
```diff
@@ -16,8 +16,8 @@
 )
 MESSAGE_CONTENT_NOTICE = "Here's the message you tried to send:"
 COPY_TEXT_HINT = (
-    "-# **Hint:** You can get your original message with formatting preserved "
-    'by using the "Copy Text" action on right-click.'
+    "-# **Hint:** you can get your original message with formatting preserved "
```
00-kat, line 19: This was lower-cased for consistency with the `TODO:`-style comments (see https://github.com/ghostty-org/discord-bot/pull/228#discussion_r2072597948).

tale (changes requested):
Look who let me in here 💀. Anyways if we're being pedantic, this should also be changed to reflect language that's mobile friendly and is probably enough in scope of the PR to be changed.

https://github.com/ghostty-org/discord-bot/blob/5f732daf44863e32aae378c895bc24772d200366/app/components/zig_codeblocks.py#L161

trag1c:
> Look who let me in here 💀. Anyways if we're being pedantic, this should also be changed to reflect language that's mobile friendly and is probably enough in scope of the PR to be changed.
> 
> https://github.com/ghostty-org/discord-bot/blob/5f732daf44863e32aae378c895bc24772d200366/app/components/zig_codeblocks.py#L161

It's worth noting that the "View whole file" option is not available on mobile at all, one has to download the file instead.

00-kat:
Mobile doesn't have syntax highlighting anyway, as far as I know.

00-kat:
@tale what's your suggestion for that one?  This is the only thing I can think of:
```diff
-    file_highlight_note = 'Click "View whole file" to see the highlighting.' 
+    file_highlight_note = 'On desktop, click "View whole file" to see the highlighting.'
```
But I don't see any way to clarify the situation on mobile without making it too long…

tale (approved):

00-kat:
…I hadn't actually changed it yet, that force push was unrelated (marking as draft so trag1c doesn't merge it pre-emptively).

trag1c (approved):

PR #248, opened 2025-06-22
Subclass `WebhookMessage` as `MovedMessage`

00-kat:
As discussed in https://github.com/ghostty-org/discord-bot/pull/245#discussion_r2159973019; I'm PRing it separately to avoid blowing up that diff.

The renaming of `_SubText` to `_Subtext` is also here because of the ordering of the commits, sorry about that.


00-kat:
Sorry for all the force-pushes, I keep discovering random things as I try work on the editing branch 😅.

PR #250, opened 2025-06-22
include more details in entity mentions

trag1c:


app/components/github_integration/mentions/discussions.py
```diff
@@ -27,4 +33,5 @@ async def get_discussion(org: str, name: str, number: int) -> Discussion:
         DISCUSSION_QUERY, variables={"number": number, "org": org, "repo": name}
     )
     data = resp["repository"]["discussion"]
+    data["answered_by"] = (answer := data.pop("answer")) and answer["user"]
```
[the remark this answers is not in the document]
trag1c: These are not functionally equivalent.

app/components/github_integration/mentions/fmt.py
```diff
@@ -58,31 +59,60 @@ def get_entity_emoji(entity: Entity) -> discord.Emoji | None:
...
-        f" in [`{owner}/{name}`](<{domain}/{owner}/{name}>)"
+        f"-# by {_format_user_link(entity.user.name)}"
+        f" in [`{owner}/{name}`](<{GITHUB_URL}/{owner}/{name}>)"
         f" on {fmt_ts('D')} ({fmt_ts('R')})\n"
     )
+    entity_detail = _format_entity_detail(entity)
 
     emoji = get_entity_emoji(entity) or "❓"
-    return f"{emoji} {headline}\n{subtext}"
+    return f"{emoji} {headline}\n{subtext}{entity_detail}"
```
00-kat, line 115: Is there any reason to put the newline in the subtext instead of the f-string at the end?

I.e.:
```diff
         f"-# by {_format_user_link(entity.user.name)}"
         f" in [`{owner}/{name}`](<{GITHUB_URL}/{owner}/{name}>)"
-        f" on {fmt_ts('D')} ({fmt_ts('R')})\n"
+        f" on {fmt_ts('D')} ({fmt_ts('R')})"
     )
     entity_detail = _format_entity_detail(entity)
 
     emoji = get_entity_emoji(entity) or "❓"
-    return f"{emoji} {headline}\n{subtext}{entity_detail}"
+    return f"{emoji} {headline}\n{subtext}\n{entity_detail}"
```
(GitHub won't let me add a suggestion so ` ```diff ` it is…)

app/components/github_integration/models.py
```diff
@@ -68,16 +69,25 @@ def kind(self) -> str:
 class Issue(Entity):
     closed: Annotated[bool, Field(alias="state"), BeforeValidator(state_validator)]
     state_reason: Literal["completed", "reopened", "not_planned", "duplicate"] | None
+    labels: list[str]
+
+    @field_validator("labels", mode="before")
+    @classmethod
+    def capitalize(cls, value: list[IssuePropLabelsItemsOneof1]) -> list[str]:
+        return [cast("str", label.name) for label in value]
```
00-kat, line 77: …that's not capitalisation though?

00-kat (approved):

app/components/github_integration/mentions/fmt.py
trag1c: Really just consistency with `_format_entity_detail` which puts a `\n` by itself (and that one's needed there to not duplicate a newline when there's no detail to be shown).

app/components/github_integration/mentions/fmt.py
```diff
@@ -14,6 +14,7 @@
     import discord
 
 ENTITY_TEMPLATE = "**{entity.kind} [#{entity.number}](<{entity.html_url}>):** {title}"
+GITHUB_URL = "https://github.com"
```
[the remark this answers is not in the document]
trag1c: > Is there any benefit to hardcoding it?

Not really. I'll revert that and just use the `domain` var and a string literal where relevant ^^

app/components/github_integration/models.py
00-kat: I coincidentally noticed that the example in Pydantic's documentation for validators used `capitalize` an hour ago 😆.

PR #258, opened 2025-06-26
Allow editing moved messages

00-kat:
Closes section three of #155.  Based on #257; merge that first.

The diff is so bad I want to give GitHub an award because how do you even mess up this bad 💀.

### DIFF
--- a/app/components/move_message.py
+++ b/app/components/move_message.py
@@ -295,7 +295,7 @@ class Ghostping(discord.ui.View):
 
 
 class HelpPostTitle(discord.ui.Modal, title="Turn into #help post"):
-    title_: discord.ui.TextInput[Self] = discord.ui.TextInput(
+    title_ = discord.ui.TextInput[Self](
         label="#help post title", style=discord.TextStyle.short, max_length=100
     )
 
@@ -487,7 +487,7 @@ class ChooseMessageAction(discord.ui.View):
 
 
 class EditMessage(discord.ui.Modal, title="Edit Message"):
-    new_text: discord.ui.TextInput[Self] = discord.ui.TextInput(
+    new_text = discord.ui.TextInput[Self](
         label="New message content",
         style=discord.TextStyle.long,
         default=discord.utils.MISSING,


app/components/move_message.py
```diff
@@ -1,25 +1,231 @@
...
+    # signals to us that the user opted to continue.
+    elapsed = dt.datetime.now(tz=dt.UTC) - edit_threads[thread.id].last_update
+    if elapsed >= dt.timedelta(minutes=15):
+        # Fifteen minutes (time_to_warning's delta + remaining) have passed, so
+        # we shall delete the thread now.
+        await _remove_edit_thread(thread, author, action="abandoned editing of")
+        return
+    # Restart the wait time. The start of this function will deal with the
+    # remaining time left to sleep.
+    await _remove_edit_thread_after_timeout(thread, author)
```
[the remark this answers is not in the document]
00-kat: I didn't want to indent the entire function, and there's no way this will ever hit the recursion max (unless someone presses continue every fifteen minutes for a week).

app/components/move_message.py
```diff
@@ -124,28 +326,335 @@ async def on_submit(self, interaction: discord.Interaction) -> None:
...
-        style=discord.ButtonStyle.danger,
+    async def show_help(self, interaction: discord.Interaction) -> None:
+        self.help_button.disabled = True
+        await interaction.response.edit_message(
+            content=f"{MESSAGE_EDIT_HELP}\n\n{EDIT_METHOD_PROMPT}", view=self
+        )
+
+
+class EditMessage(discord.ui.Modal, title="Edit Message"):
+    new_text: discord.ui.TextInput[Self] = discord.ui.TextInput(
```
trag1c, line 492:
```suggestion
    new_text = discord.ui.TextInput[Self](
```
`HelpPostTitle` (albeit not touched in this PR) should get the same treatment.

app/components/move_message.py
```diff
@@ -124,28 +326,335 @@ async def on_submit(self, interaction: discord.Interaction) -> None:
...
-        self,
-        interaction: discord.Interaction,
-        button: discord.ui.Button[Self],
+
+    def __init__(self, message: MovedMessage, split_subtext: SplitSubtext) -> None:
+        self._split_subtext = split_subtext
+        self.new_text.default = split_subtext.content
+        # Subtract one to account for the newline character.
+        self.new_text.max_length = 2000 - len(split_subtext.subtext) - 1
+        super().__init__()
```
trag1c, line 503: Any reason this `super` call is not the first thing done in the `__init__`? Feels off to me.

app/components/move_message.py
```diff
@@ -1,25 +1,231 @@
...
+    "all other cases."
+)
+ALREADY_EDITING = (
+    "You are already editing a message in a thread! You can only edit one "
+    "message in a thread at a time, to prevent abuse. Please finish editing "
+    "the message at <#{thread_id}> first, or use the modal."
+)
+NO_THREAD_PERMS = (
+    "⚠️ I don't have the required permissions to create private "  # test: allow-vs16
+    "threads; please contact a moderator! In the meantime, use the modal "
```
trag1c, line 66: Wondering if this should say `moderator` or `administrator` (or maybe both), since only admins can fix those kinds of issues. However, it's not instantly obvious who is an admin + I suppose moderators can just forward any issues to admins 😛

app/components/move_message.py
```diff
@@ -1,25 +1,231 @@
...
+    "modal instead, or try again later."
+)
+NO_CONTENT_TO_EDIT = (
+    "{}, your message does not currently have any content. Sending a message "
+    "would add the specified content to the message. Any attachments you "
+    "upload are also added to your message."
+)
+EDIT_IN_THREAD_HINT = (
+    "Please send a message containing the message's new content.\n"
+    "-# **Hint:** you can copy your message's content by pressing "
```
trag1c, line 80: Is it just me or does "pressing" sound more computer-centric?

app/utils/message_data.py
```diff
@@ -50,13 +47,7 @@ async def scrape(cls, message: discord.Message) -> Self:
...
-        for file in attachments:
-            if file.size > MAX_ATTACHMENT_SIZE:
-                skipped_attachments += 1
-                continue
-            fp = BytesIO(await file.read())
-            files.append(discord.File(fp, filename=file.filename))
-        self.files = files
-        self.skipped_attachments = skipped_attachments
+        self.files = await asyncio.gather(
+            *(a.to_file() for a in attachments if a.size <= MAX_ATTACHMENT_SIZE)
```
trag1c, line 51: Great find!

app/utils/message_data.py
```diff
@@ -39,9 +39,6 @@ def __init__(self, message: discord.Message) -> None:
 
 
 class MessageData(ExtensibleMessage):
-    files: list[discord.File]
-    skipped_attachments: int
-
```
trag1c, line 44: Why were these removed?

app/components/move_message.py
```diff
@@ -1,25 +1,231 @@
...
+                *(await channel.fetch_message(moved_message.id)).attachments,
+                *(await MessageData.scrape(message)).files,
+            ],
+            allowed_mentions=discord.AllowedMentions.none(),
+        )
+
+
+async def _remove_edit_thread(
+    thread: discord.Thread,
+    author: discord.User | discord.Member,
```
trag1c, line 165: Use `app.utils.Account` instead. Also applies to `_remove_edit_thread_after_timeout`.

app/components/move_message.py
```diff
@@ -1,25 +1,231 @@
...
+    *,
+    action: str = "finished editing",
```
trag1c, line 167: I think there should be no default value and `"finished editing"` should just be explicitly stated whenever the case. It doesn't seem any more special than other actions to me.

app/components/move_message.py
```diff
@@ -124,28 +326,335 @@ async def on_submit(self, interaction: discord.Interaction) -> None:
...
-        self.message = message
+        self._message = message
+        self._split_subtext = SplitSubtext(message)
+        self._attachment_button_added = False
+        # Adding the thread button might also add the attachment button so that
+        # the attachment button can be placed in between the thread and help
+        # buttons.
+        self._add_thread_button()
+        if not self._attachment_button_added:
+            self._add_attachment_button()
```
trag1c, line 344: `_add_thread_button` doesn't modify `_attachment_button_added`, which currently makes the variable useless if I'm reading this right.

app/utils/__init__.py
```diff
@@ -106,6 +103,22 @@ async def delete(
         )
 
 
+class DeleteInstead(discord.ui.View):
```
trag1c, line 106: This should be added to `__all__`.

trag1c (changes requested):

app/components/move_message.py
trag1c: Right, I missed that!

app/components/move_message.py
00-kat: When I initially added this function, there was one failure case and every other case was `"finished editing"`, so I decided to make that the default, though grepping through now shows that it's not particularly special anymore.

app/utils/message_data.py
00-kat: If I remember correctly, they were there because I initially did something like this:
```py
class MessageData(...):
    files: list[discord.File]
    skipped_attachments: int
    async def _get_files(...):
        self.files = []
        ...
```
to avoid adding a type annotation to `self.files = []`.  However, in a code review you asked me to use a local and set `self.files` right at the end, which means the type annotation is required again, except I then forgot to remove those two from the top of the class and only noticed it now when changing it.

app/components/move_message.py
00-kat: I'm the opposite: “pressing” always gives me touchscreen vibes, so I generally think it's more mobile-centric, lol.  I chose pressing since it's the name of the action you do to a physical button with your finger, rather than an action you do to your mouse (clicking) or an action you do to your screen (tapping).

app/components/move_message.py
```diff
@@ -124,28 +326,335 @@ async def on_submit(self, interaction: discord.Interaction) -> None:
...
+                auto_archive_duration=60,
+                reason=f"{interaction.user.name} wants to edit a moved message",
+                invitable=False,
+            )
+        except discord.Forbidden:
+            self.thread_button.disabled = True
+            await interaction.response.edit_message(content=NO_THREAD_PERMS, view=self)
+            raise  # Also log it in Sentry.
+        except discord.HTTPException as e:
+            if e.code != MAXIMUM_NUMBER_OF_ACTIVE_THREADS_REACHED:
```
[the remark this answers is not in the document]
00-kat: No clue, I just went through the entire error list and the two handled here are the only theoretical possibilities.  It probably was possible in the past before the autoclose task was added, but I doubt it these days.

app/components/move_message.py
00-kat: When I was writing it I was debating the exact same thing, and

> However, it's not instantly obvious who is an admin + I suppose moderators can just forward any issues to admins 😛

I also reached the exact same conclusion before deciding to leave it as moderators 😅.  Most people don't really know who's an admin and who's a moderator.

app/components/move_message.py
00-kat: Huh, I remember trying this when I initially changed everything to use `Self` in another PR and either Ruff or Python gave a syntax error 🤔.

PR #268, opened 2025-07-08
Add tests for assorted utils

### LINKED ISSUE #265: Add tests for assorted utils
Most of these shouldn't be too annoying to test without extensive mocking of discord.py.

00-kat:
Closes #265.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,9 @@ pylint.max-branches = 20
 pylint.max-returns = 15
 isort.known-local-folder = ["app"]
 
+[tool.ruff.lint.per-file-ignores]
+"tests/*" = ["FBT"]
+
 [tool.pyright]
 typeCheckingMode = "strict"
 reportUnknownLambdaType = false
--- a/tests/test_message_moving.py
+++ b/tests/test_message_moving.py
@@ -34,7 +34,7 @@ def test_format_unicode_partialemoji(emoji: str) -> None:
         ("fooBaRb__az_loR_em", False, 1328977544984268890),
     ],
 )
-def test_format_partial_emoji(*, name: str, animated: bool, id_: int) -> None:
+def test_format_partial_emoji(name: str, animated: bool, id_: int) -> None:
     url = f"https://cdn.discordapp.com/emojis/{id_}.{'gif' if animated else 'png'}"
     assert (
         _format_emoji(discord.PartialEmoji(name=name, animated=animated, id=id_))
@@ -45,7 +45,7 @@ def test_format_partial_emoji(*, name: str, animated: bool, id_: int) -> None:
 @pytest.mark.parametrize(
     ("is_usable", "output"), [(True, "<foo>"), (False, "[foo](<bar>)")]
 )
-def test_format_emoji_is_usable(*, is_usable: bool, output: str) -> None:
+def test_format_emoji_is_usable(is_usable: bool, output: str) -> None:
     fake_emoji = Mock(
         discord.Emoji,
         is_usable=Mock(return_value=is_usable),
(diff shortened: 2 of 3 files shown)


app/utils/__init__.py
```diff
@@ -187,7 +187,7 @@ def is_attachment_only(
 ) -> bool:
     if preprocessed_content is None:
         preprocessed_content = message.content
-    return not any((
+    return bool(message.attachments) and not any((
```
trag1c, line 190: What's the reason for this change? My first reaction is that it's only needed to make tests pass but that sounds sus. Was there a bug that this fixes?

tests/fixtures/config.py
```diff
@@ -2,35 +2,36 @@
...
+    "BOT_HELPER_ROLE_ID": "070411150417",
+}
 
 
 @pytest.fixture
 def bot_env() -> SimpleNamespace:
-    return BOT_ENV
+    return SimpleNamespace(**{
+        k.removeprefix("BOT_") if k != "BOT_TOKEN" else k: v for k, v in BOT_ENV.items()
+    })
```
trag1c, line 31: Why is `BOT_TOKEN` an exception? Also, I assume you rewrote `BOT_ENV` to create `SimpleNamespace` in the fixture to prevent global state modification, correct? :)

trag1c (commented):
Haven't looked at the actual tests yet, just wanted to ask about some things I found confusing on my initial skim.

app/utils/__init__.py
00-kat: I randomly noticed in my initial inspection of the source code that a message with nothing in it may be counted as “having attachments only” so decided to change it; it doesn't fix any bug and the test for this case was only added a while after I made the change.

tests/fixtures/config.py
00-kat: I have no clue, it's the only exception in `config.py`: https://github.com/ghostty-org/discord-bot/blob/f06a3028e5cb1b925bdba53c0e35b0601d78166b/app/config.py#L7
So I match the behaviour here by once again making it the only exception.

> Also, I assume you rewrote `BOT_ENV` to create `SimpleNamespace` in the fixture to prevent global state modification, correct? :)

I'm not sure what you mean here, `SimpleNamespace`s are mutable too and I never reassign `BOT_ENV`. If you're asking why I convert it to a `SimpleNamespace` here, that would be because this namespace is monkeypatched into the `config` module, and all users access `config.FOO` and not `config["FOO"]`, so leaving it a dict would break every test using this fixture.

trag1c: I was referring to this:
```py
FOO = SimpleNamespace(a=1, b=2)
BAR = {"a": 1, "b": 2}

@pytest.fixture
def foo() -> SimpleNamespace:
    return FOO

@pytest.fixture
def bar() -> SimpleNamespace:
    return SimpleNamespace(**BAR)

def test_foo_bar(foo, bar):
    assert foo.a == bar.a
    foo.a = 0
    bar.a = 0

def test_bar_foo(foo, bar):
    assert foo.a == bar.a
```
The second test fails because `foo.a = 0` modifies the global `FOO`, while `bar.a = 0` only modifiers the newly created namespace, so it's trying to `assert 0 == 1`

00-kat: Ah, I thought they were re-instantiated for every test 😅.

trag1c: No, other than the call being triggered by being listed as a test param, fixtures (to my knowledge) aren't really different from regular function calls so both of these just return references.

> I have no clue, it's the only exception in `config.py`

Oh yeah true lol.

PR #271, opened 2025-07-11
rework the hook system

### LINKED ISSUE #148: Allow freezing entity mentions
This would stop edit/deletion reactions and let the author do whatever.

### LINKED ISSUE #185: zig-codeblocks delete hook crashes w/ multiple msg processors
Not 100% sure the zig-codeblocks delete hook is at fault here but I have pretty good reason to believe so.

I can reproduce this when mixing entity mentions with Zig codeblocks but not when mixing e.g. entity mentions with code links.

## To reproduce (sometimes)
1. Send a message like:
`````
#1234 ```zig
const std = @import("std");```
`````
2. Delete that message
3. The hook might have crashed

## Example traceback
```
KeyError -> <Message id=1350594636623773787 channel=<TextChannel id=1281626115198292030 name='development' position=8 nsfw=False news=False category_id=1281625995924865035> type=<MessageType.default: 0> author=<Member id=252206453878685697 name='trag1c' global_name='trag1c' bot=False nick=None guild=<Guild id=1281624934636064811 name='Ghostty Bot Testing' shard_id=0 chunked=True member_count=17>> flags=<MessageFlags value=0>>
  File "/Users/trag1c/dev/download/ghostty-bot/.venv/lib/python3.12/site-packages/discord/client.py", line 481, in _run_event
    await coro(*args, **kwargs)
  File "/Users/trag1c/dev/download/ghostty-bot/app/core.py", line 114, in on_message_delete
    await zig_codeblock_delete_hook(message)
  File "/Users/trag1c/dev/download/ghostty-bot/app/components/zig_codeblocks.py", line 286, in zig_codeblock_delete_hook
    codeblock_linker.unlink(message)
  File "/Users/trag1c/dev/download/ghostty-bot/app/utils/hooks.py", line 31, in unlink
    del self._refs[original]
        ~~~~~~~~~~^^^^^^^^^^
```

### PR DESCRIPTION
**Pull request by trag1c**
Closes #148. Fixes #185.

### CODE REVIEW

File: app/utils/hooks.py [hunk 1]
```diff
@@ -1,12 +1,20 @@
...
 
 import discord
 
 
+@dataclass(frozen=True, slots=True, kw_only=True)
+class ProcessedMessage:
+    item_count: int
+    content: str = ""
+    files: list[discord.File] = field(default_factory=list[discord.File])
+    embeds: list[discord.Embed] = field(default_factory=list[discord.Embed])
```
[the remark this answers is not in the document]
trag1c: Yeah, same.

00-kat (changes requested):

app/components/zig_codeblocks.py
```diff
@@ -146,173 +97,58 @@ async def _prepare_reply(
...
-    file_highlight_note = 'On desktop, click "View whole file" to see the highlighting.'
-    if codeblocks and any(block.lang == "zig" for block in codeblocks):
-        zig_code = process_discord_markdown(message.content, only_code=True)
-        zig_code += f"\n{file_highlight_note}" if attachments else ""
-        return _split_codeblocks(zig_code), attachments
+    if (
+        codeblocks
+        and len(message.content) <= 2000
+        and any(block.lang == "zig" for block in codeblocks)
+    ):
```
00-kat, line 104: This loop runs for code blocks that aren't Zig code blocks, turning something like
````md
```zig
foo bar baz
``````rs
foo rar raz
```
````
into
````md
```ansi
foo bar baz```

```ansi
foo rar raz```
````


```suggestion
    codeblocks = [c for c in extract_codeblocks(message.content) if c.lang == "zig"]
    if codeblocks and len(message.content) <= 2000:
```

00-kat: I have no clue why that suggestion includes the removed lines too; for clarity, this is what the diff looks like:
```diff
-    codeblocks = list(extract_codeblocks(message.content))
+    codeblocks = [c for c in extract_codeblocks(message.content) if c.lang == "zig"]
-    if (
-        codeblocks
-        and len(message.content) <= 2000
-        and any(block.lang == "zig" for block in codeblocks)
-    ):
+    if codeblocks and len(message.content) <= 2000:
```

app/components/zig_codeblocks.py
```diff
@@ -146,173 +97,58 @@ async def _prepare_reply(
...
 
     codeblocks = list(extract_codeblocks(message.content))
-    file_highlight_note = 'On desktop, click "View whole file" to see the highlighting.'
-    if codeblocks and any(block.lang == "zig" for block in codeblocks):
-        zig_code = process_discord_markdown(message.content, only_code=True)
-        zig_code += f"\n{file_highlight_note}" if attachments else ""
-        return _split_codeblocks(zig_code), attachments
+    if (
+        codeblocks
+        and len(message.content) <= 2000
```
00-kat, line 102: Hold on why do we have this check anyway? It would mean a 3812 character message with a 60 character code block within wouldn't be responded to, for example. And it wouldn't really affect any of the code below; it's not possible for the user to get the bot to attach over ten files except when attaching Zig files for highlighting (which wouldn't be stopped by this condition) even if https://github.com/ghostty-org/discord-bot/pull/271#discussion_r2200490149 is fixed by sorting the code blocks by size while popping (see below for the reason), and I don't see anything else that this check could guard against. The “replace my message” button's code handles this case fine too.

```suggestion
```

As for why it's not possible to go over the attachment limit if every attachment contains over 2000 characters and the user doesn't attach files:

The highest original-to-highlighted character ratio is with `0-` (or any other number and any other character that isn't highlighted the same as the number), which is a 2:11 ratio, as the highlighted output is `[36m0[0m-`. This can be repeated infinitely too, making the smallest code block that gets attached as a file this:
````md
```zig
0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-```
````
Removing even a single other character results in the code block being included in the message content instead of attached as a file.

This code block (including the Markdown syntax) is 372 characters long, which means non-Nitro users can send at most five code blocks that get attached as a file (adding a sixth one of these goes 232 over the limit), and Nitro users can send at most ten of these (adding a eleventh goes 92 over the limit; if we didn't need the ` ```zig` and ` ``` `, eleven would fit in 4000 characters).

Discord's attachment limit is 10, and since the maximum theoretically possible (unless the theme is changed, or highlighting is added for more symbols (such as `?` and `*`) in `zig-codeblocks`) amount of code blocks is also 10, there won't be too many attachments (unless, again, files are uploaded by the user themself, in which case this condition won't do anything anyway).

File: app/components/zig_codeblocks.py [hunk 4]
```diff
@@ -121,22 +74,20 @@ async def replace(
...
     attachments: list[discord.File] = []
     for att in message.attachments:
         if not att.filename.endswith(".zig") or att.size > MAX_ZIG_FILE_SIZE:
             continue
         content = (await att.read())[:MAX_CONTENT]
         if content.count(b"\n") <= 5 and len(content) <= 1900:
-            message.content = f"```zig\n{content.decode()}```\n{message.content}"
+            message.content = f"{CodeBlock('zig', content.decode())}\n{message.content}"
             continue
         attachments.append(
```
00-kat, line 92: I entirely missed this as well, but here's *yet another* way to add too many attachments; this one defies basically every countermeasure except skipping a few code blocks.

Send a message with this content:
````md
```zig
0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-```
````
And attach this file ten times: [lots of code.zig.txt](https://github.com/user-attachments/files/21192483/lots.of.code.zig.txt). (Note: GitHub doesn't allow uploading `.zig` files, so remove the `.txt` suffix before uploading.)

This will result in the bot trying to attach eleven attachments. You can even pair ten attachments alongside \<see edit\> to get the bot to attempt sending twenty attachments.

**Review by 00-kat — CHANGES_REQUESTED** [review]

File: app/components/zig_codeblocks.py [hunk 5]
```diff
@@ -34,81 +29,39 @@
...
+
+class CodeblockActions(ItemActions):
+    linker = codeblock_linker
+    action_singular = "sent this code block"
+    action_plural = "sent these code blocks"
+
+    def __init__(self, message: discord.Message, item_count: int) -> None:
+        super().__init__(message, item_count)
+        replaced_content = apply_discord_wa(process_markdown(message.content, THEME))
+        if message.attachments or len(replaced_content) > 2000:
```
00-kat, line 55: Why is replacing a message that contains attachments disallowed?

app/utils/hooks.py
```diff
@@ -18,99 +26,106 @@ async def remove_view_after_timeout(
...
+        if linker.is_frozen(before):
             return
 
-        content, files, embeds = extract_content(content)
-        if not (content or files or embeds):
-            # The message is empty, don't send a message with only a view
+        # Some processors use negative values to symbolize special error values, so this
+        # can't be `== 0`. An example of this is the snippet_message() function in the
+        # file app/components/github_integration/code_links.py.
+        if new_output.item_count <= 0 and not linker.is_frozen(before):
```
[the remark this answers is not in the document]
trag1c: Good catch!

app/core.py
```diff
@@ -111,20 +111,24 @@ async def on_message(message: discord.Message) -> None:
 
 @bot.event
 async def on_message_edit(before: discord.Message, after: discord.Message) -> None:
-    await entity_mention_edit_hook(before, after)
```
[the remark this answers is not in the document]
trag1c: *I think* it's because unlinking was unsafe. It's now idempotent so it's safe to do this.

app/components/zig_codeblocks.py
```diff
@@ -146,173 +97,58 @@ async def _prepare_reply(
...
+    ):
+        highlighted_codeblocks = [
+            CodeBlock("ansi", apply_discord_wa(highlight_zig_code(c.body, THEME)))
+            for c in codeblocks
+        ]
+        max_length = 2000 - (len(FILE_HIGHLIGHT_NOTE) - 1 if attachments else 0)
+        while len(code := "\n\n".join(map(str, highlighted_codeblocks))) > max_length:
+            file = discord.File(
+                BytesIO(highlighted_codeblocks.pop().body.encode()),
+                filename=f"{len(highlighted_codeblocks)}.ansi",
```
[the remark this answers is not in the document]
trag1c: That was intended.

app/core.py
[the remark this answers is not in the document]
00-kat: That was my guess too but it seemed like a rather simple case to fix so I wasn't sure lol.

app/components/zig_codeblocks.py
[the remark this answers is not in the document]
00-kat: My opinion is that one-based indexing is a better idea for UI-related things (since we normally start counting from 1) and these file names are rather prominently displayed in the UI.

It doesn't matter too much though since mostly only programmers would see them and it wouldn't be too difficult to realise that it starts from 0 just by looking at the first file.

app/components/zig_codeblocks.py
```diff
@@ -146,173 +97,58 @@ async def _prepare_reply(
...
+        and any(block.lang == "zig" for block in codeblocks)
+    ):
+        highlighted_codeblocks = [
+            CodeBlock("ansi", apply_discord_wa(highlight_zig_code(c.body, THEME)))
+            for c in codeblocks
+        ]
+        max_length = 2000 - (len(FILE_HIGHLIGHT_NOTE) - 1 if attachments else 0)
+        while len(code := "\n\n".join(map(str, highlighted_codeblocks))) > max_length:
+            file = discord.File(
+                BytesIO(highlighted_codeblocks.pop().body.encode()),
```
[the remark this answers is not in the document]
trag1c: Outdated, since code blocks are now chosen based on height first.

app/components/zig_codeblocks.py
trag1c: My guess would be laziness in supporting that lol.

app/utils/hooks.py
```diff
@@ -18,99 +26,106 @@ async def remove_view_after_timeout(
...
         return next(
-            (msg for msg, replies in self._refs.items() if reply in replies), None
+            (msg for msg, reply_ in self._refs.items() if reply == reply_), None
         )
 
     def unlink_from_reply(self, reply: discord.Message) -> None:
         if (original_message := self.get_original_message(reply)) is not None:
             self.unlink(original_message)
 
-    def unlink_if_expired(self, reply: discord.Message) -> bool:
```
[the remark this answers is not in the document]
trag1c: How's "do a cleanup on new link"? So basically:
```py
class MessageLinker:
    def link(self, original: discord.Message, reply: discord.Message) -> None:
        self._free_dangling_links()
        # *the actual linking after*
```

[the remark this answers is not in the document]
00-kat: I think that's a better idea than what I had thought yeah.

app/components/zig_codeblocks.py
```diff
@@ -121,198 +81,123 @@ async def replace(
...
-            files=files,
-            mention_author=False,
-        )
-        codeblock_linker.link(message, reply)
-        await remove_view_after_timeout(reply, VIEW_TIMEOUT)
-        return
+def _add_user_notes(
+    content: str, omitted_codeblocks: int, attachments: Collection[discord.File]
+) -> str:
+    omission_note = "\n-# {} codeblock{} omitted"
```
00-kat, line 123: Is there any reason this isn't a constant? Or why not just use an f-string at the one place this is used?

app/components/zig_codeblocks.py
```diff
@@ -121,198 +81,123 @@ async def replace(
...
-            # There was no code before, so treat this as a new message.
-            # Note: No new attachments can appear, their count can only decrease.
-            await check_for_zig_code(after)
-        # The message was removed from the M2C map at some point
-        return
+        if not omitted_codeblocks:
+            if max_length == 2000:
+                # We definitely have attachments at this point
+                max_length -= len(FILE_HIGHLIGHT_NOTE)
+            max_length -= 30  # expected omission note size (conservative)
```
00-kat, line 168: If the omission note template is a const this could be something like

```suggestion
            max_length -= len(OMISSION_NOTE) + 5  # expected omission note size (conservative)
```

app/components/zig_codeblocks.py
```diff
@@ -121,198 +81,123 @@ async def replace(
...
-        if not (old_contents or before in frozen_messages):
-            # There was no code before, so treat this as a new message.
-            # Note: No new attachments can appear, their count can only decrease.
-            await check_for_zig_code(after)
-        # The message was removed from the M2C map at some point
-        return
+        if not omitted_codeblocks:
+            if max_length == 2000:
+                # We definitely have attachments at this point
+                max_length -= len(FILE_HIGHLIGHT_NOTE)
```
00-kat, line 167: Shouldn't this be done above when `attachments.append()` is done? I.e.:
```py
if len(attachments) < 10:
    if not attachments:
        # This is the first attachment
        max_length -= len(FILE_HIGHLIGHT_NOTE)
    attachments.append(file)
    continue
```

00-kat (changes requested):

app/components/zig_codeblocks.py
trag1c: You're right, good catch!

app/components/zig_codeblocks.py
trag1c: No reason (maybe there was one in the original version tho), corrected to a const!

app/components/zig_codeblocks.py
trag1c: Tracking at #277.

app/utils/hooks.py
```diff
@@ -18,99 +26,113 @@ async def remove_view_after_timeout(
...
+    def is_frozen(self, message: discord.Message) -> bool:
+        return message in self._frozen
 
-    def link(self, original: discord.Message, *replies: discord.Message) -> None:
-        self._refs[original].extend(replies)
+    def get(self, original: discord.Message) -> discord.Message | None:
+        return self._refs.get(original)
+
+    def _free_dangling_links(self) -> None:
+        # Saving keys to a tuple to avoid a "changed size during iteration" error
```
00-kat, line 49: This clones every single link right? I think instead the ones that are to be removed should be added to a new set, then those refs should be removed in one go at the end (I'm not sure if there's an equivalent of `set.difference()` for dicts though).

app/utils/hooks.py
```diff
@@ -18,99 +26,113 @@ async def remove_view_after_timeout(
...
+        for msg in tuple(self._refs):
```
00-kat, line 50: Doesn't `self._frozen` need cleaning-up too?

00-kat (changes requested):

app/utils/hooks.py
00-kat: Ah okay; disregard my comment then.

app/utils/hooks.py
trag1c: Correct, fixed!

00-kat (approved):

PR #287, opened 2025-07-26
fix: SGR 0 only applied inside codeblocks

### LINKED ISSUE #274: Replaced Zig codeblock messages have SGR 0 before *all* `// `s
Examples:
````md
hi! // hello!!
```zig
foo
```
````
````md
In python, `2 // 3` becomes 0. In Zig,
```zig
2 // 3
```
becomes 2.
````

The above examples, rendered; a black box with a question mark, then a `[0m`, precede every `//`.

### PR DESCRIPTION
**Pull request by CoderJoshDK**
Resolves #274

This PR only applies the discord work around from within the codeblocks. This is the cleanest solution I could come up with, without modifying the upstream library. 

Thing to note, I also made it only change code inside ANSI codeblocks. Don't want to do the sub on "multi codeblock" messages.

image

Example:

````md
In python `2 // 3` is 0 and 2 in zig
```py
2 // 3
```
Outputs 0

```zig
2 // 3
```
Outputs 2

hi! // hello!!

```py
# Extra disambiguous codeblock that doesn't show up in replace interaction
```
````

### DIFF
--- a/app/components/zig_codeblocks.py
+++ b/app/components/zig_codeblocks.py
@@ -58,19 +58,14 @@ def apply_discord_wa(source: str) -> str:
     return source.replace("///", "\x1b[0m///").replace("// ", "\x1b[0m// ")
 
 
-def _apply_discord_wa_in_only_codeblock(source: str) -> str:
+def _apply_discord_wa_in_ansi_codeblocks(source: str) -> str:
     # Resolves #274
     code_blocks = extract_codeblocks(source)
-    code_blocks = [  # Remove duplicates
-        v1
-        for i, v1 in enumerate(code_blocks)
-        if not any(
-            (v1.body == v2.body) and (v1.lang == v2.lang) for v2 in code_blocks[:i]
-        )
-    ]
+    # Remove duplicates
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by CoderJoshDK** [comment]
You are right. I originally didn't do that because I assumed the performance would be much slower. And it likely is,,, but this is only triggered on zig messages with less than 2000 characters and for a single server. I think it is fine this way.
Hahahah, changed it.

image

File: app/components/zig_codeblocks.py [hunk 1]
```diff
@@ -58,14 +58,33 @@ def apply_discord_wa(source: str) -> str:
...
+    code_blocks = [  # Remove duplicates
+        v1
+        for i, v1 in enumerate(code_blocks)
+        if not any(
+            (v1.body == v2.body) and (v1.lang == v2.lang) for v2 in code_blocks[:i]
+        )
+    ]
+    for block in code_blocks:
+        if block.lang == "ansi":
+            body = f"```{block.lang}\n{block.body}```"
```
trag1c, line 73:
```suggestion
            body = str(block)
```

trag1c (changes requested):

trag1c (approved):
Actually my only suggestions were simple enough I just applied them directly. Thank you!

PR #288, opened 2025-07-27
re-suppress embeds after 5 seconds

### LINKED ISSUE #273: Suppress embeds more than once
Marking this as a bug even though it's not really our fault.

This mostly happens with GitHub entity/comment/code links, where the bot attempts to block the embed before it loads, resulting in inconsistent suppression.

My current idea is to just throw in an additional suppress after a few seconds.

trag1c:
Closes #273

### DIFF
--- a/app/components/github_integration/code_links.py
+++ b/app/components/github_integration/code_links.py
@@ -17,7 +17,7 @@ from app.common.hooks import (
     ProcessedMessage,
     create_delete_hook,
     create_edit_hook,
-    remove_view_after_timeout,
+    remove_view_after_delay,
 )
 from app.components.zig_codeblocks import THEME
 from app.setup import gh
@@ -170,7 +170,7 @@ async def reply_with_code(message: dc.Message) -> None:
         view=CodeLinkActions(message, output.item_count),
     )
     code_linker.link(message, sent_message)
-    await remove_view_after_timeout(sent_message)
+    await remove_view_after_delay(sent_message)
 
 
 code_link_delete_hook = create_delete_hook(linker=code_linker)
--- a/app/components/github_integration/code_links.py
+++ b/app/components/github_integration/code_links.py
@@ -1,3 +1,4 @@
+import asyncio
 import re
 import string
 import urllib.parse
@@ -21,6 +22,7 @@ from app.common.hooks import (
 )
 from app.components.zig_codeblocks import THEME
 from app.setup import gh
+from app.utils import suppress_embeds_after_delay
 
 CODE_LINK_PATTERN = re.compile(
     r"https?://(?:www\.)?github\.com/([^/\s]+)/([^/\s]+)/blob/([^/\s]+)/([^\?#\s]+)"
@@ -170,7 +172,10 @@ async def reply_with_code(message: dc.Message) -> None:
         view=CodeLinkActions(message, output.item_count),
     )
     code_linker.link(message, sent_message)
-    await remove_view_after_delay(sent_message)
+    await asyncio.gather(
+        suppress_embeds_after_delay(message),
+        remove_view_after_delay(sent_message),
+    )
 
 
 code_link_delete_hook = create_delete_hook(linker=code_linker)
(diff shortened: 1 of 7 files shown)


app/components/github_integration/comments/integration.py
```diff
@@ -96,7 +98,10 @@ async def reply_with_comments(message: dc.Message) -> None:
     )
     await message.edit(suppress=True)
     comment_linker.link(message, sent_message)
-    await remove_view_after_timeout(sent_message)
+    await asyncio.gather(
+        remove_view_after_timeout(sent_message),
```
00-kat, line 102: It doesn't matter but I kinda think it should be the other way around because the suppress timeout runs first.

```suggestion
        suppress_embeds_after_timeout(sent_message),
        remove_view_after_timeout(sent_message),
```

00-kat, line 102: It doesn't matter but I kinda think it should be the other way around because the suppress timeout runs first.

```suggestion
        suppress_embeds_after_timeout(sent_message),
        remove_view_after_timeout(sent_message),
```

(I can't get the suggestion to work no matter how hard I try, thank you GitHub /s /nc.)

app/components/github_integration/code_links.py
```diff
@@ -170,7 +172,10 @@ async def reply_with_code(message: dc.Message) -> None:
         view=CodeLinkActions(message, output.item_count),
     )
     code_linker.link(message, sent_message)
-    await remove_view_after_timeout(sent_message)
+    await asyncio.gather(
+        remove_view_after_timeout(sent_message),
+        suppress_embeds_after_timeout(sent_message),
```
00-kat: Ditto.

```suggestion
        suppress_embeds_after_timeout(sent_message),
        remove_view_after_timeout(sent_message),
```

app/utils/__init__.py
```diff
@@ -127,3 +129,8 @@ def is_attachment_only(
         message.poll,
         message.stickers,
     ))
+
+
+async def suppress_embeds_after_timeout(message: dc.Message, delay: float = 5) -> None:
```
00-kat, line 134: `float` then `5` looks odd to me; it also doesn't match `remove_view_after_timeout()`.

```suggestion
async def suppress_embeds_after_timeout(message: dc.Message, delay: float = 5.0) -> None:
```

app/components/github_integration/mentions/resolution.py
```diff
@@ -98,3 +101,6 @@ async def resolve_repo_signatures(
         valid_signatures += 1
         if valid_signatures == 10:
             break
+
+    if suppress_again:
+        await suppress_embeds_after_timeout(message)
```
00-kat, line 106: This stops the coro for five seconds so the bot doesn't respond with the mentions until after this timeout elapses.

app/components/github_integration/comments/integration.py
00-kat: This suggestion is outdated but GitHub doesn't say so for some reason.

app/components/github_integration/mentions/integration.py
```diff
@@ -50,7 +54,16 @@ async def reply_with_entities(message: dc.Message) -> None:
         view=MentionActions(message, output.item_count),
     )
     mention_linker.link(message, sent_message)
-    await remove_view_after_timeout(sent_message)
+
+    coros = [remove_view_after_delay(sent_message)]
+    # The suppress is done here (instead of resolve_repo_signatures) to prevent blocking
```
00-kat, line 59:
```suggestion
    # The suppress is done here (instead of in resolve_repo_signatures) to prevent blocking
```

app/utils/__init__.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import asyncio
+import contextlib
```
00-kat, line 4: We import with `from` everywhere else.

```suggestion
from contextlib import suppress
```

00-kat (approved):

PR #300, opened 2025-08-14
Display diff size in commit mentions

### LINKED ISSUE #295: Display diff size in commit mentions
Just like PR mentions do :)
* additions/deletions can be found in `obj.stats` (the `obj` in `CommitCache._fetch`)
* you can get the changed file count by checking `len(obj.files)`
* remember not to display this if the diff is empty

00-kat:
Closes #295.


app/utils/__init__.py
```diff
@@ -138,3 +138,13 @@ async def suppress_embeds_after_delay(message: dc.Message, delay: float = 5.0) -
     await asyncio.sleep(delay)
     with safe_edit:
         await message.edit(suppress=True)
+
+
+def format_diff_note(
+    additions: int,
+    deletions: int,
+    changed_files: int | None = None,
```
trag1c, line 146: Why is the ` | None = None` part there? I don't see any case of it being possibly `None` or a call not receiving that argument at all.

00-kat: Oh oops, that was something from the past and I completely missed that when changing the function.

trag1c (approved):

PR #320, opened 2025-08-26
feat: allow for explicit definition of ghostty guild through id

### LINKED ISSUE #318: Local testing is impossible if no server with "ghostty" in the name
Currently, the code tries to find the ghostty server. Primarily for nitro emoji reasons. However, if the local bot is not in any server with "ghostty" in its title, the whole thing is impossible to test. 

The function that fetches the ghostty server can either be updated to not raise an error. Or this behavior can be noted in the README if expected.

Please weigh in and let me know the preference here.

CoderJoshDK:
Closes #318 by introducing `BOT_GUILD_ID` into the `.env`. This allows local testing to be done on a server that does not contain the phrase `ghostty` in it. If the id doesn't find a guild, the fallback is to search for the first `ghostty` server. And default to the first guild (instead of raising an error).

After writing the solution that was discussed in the issue, I think we can get away with just defaulting to the first guild and not adding an explicit environment variable. However, semantically, I think being able to disambiguate the target server is a desirable outcome.


app/common/message_moving.py
```diff
@@ -38,11 +38,7 @@
 
 
 def get_ghostty_guild() -> dc.Guild:
-    try:
-        return next(g for g in bot.guilds if "ghostty" in g.name.casefold())
-    except StopIteration:
-        msg = "bot guild name does not contain 'ghostty'"
-        raise ValueError(msg) from None
+    return bot.get_guild(config.guild_id) or bot.guilds[0]
```
00-kat, line 41: I think there should be a warning logged.

```suggestion
    if guild := bot.get_guild(config.guild_id):
        return guild
    logger.warning("guild {} not found; defaulting to first guild", config.guild_id)
    return bot.guilds[0]
```

app/setup.py
```diff
@@ -49,6 +49,7 @@ class Config(BaseSettings):
 
     help_channel_tag_ids: dict[str, int]
 
+    guild_id: int
```
00-kat, line 52: I think you should do something similar to `cache_channel` and store the guild in the config class itself.

00-kat (changes requested):

app/common/message_moving.py
trag1c: If anything, this should be a debug/info log. It's completely fine not to use the env var if the bot is only in one server.

app/setup.py
```diff
@@ -71,6 +72,13 @@ def parse_tag_ids(cls, value: str) -> dict[str, int]:
             for name, id_ in (pair.split(":") for pair in value.split(","))
         }
 
+    @cached_property
+    def ghostty_guild(self) -> dc.Guild:
+        if self.guild_id and (guild := bot.get_guild(self.guild_id)):
+            return guild
+        logger.info("guild {} not found; defaulting to first guild", self.guild_id)
```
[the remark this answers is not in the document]
CoderJoshDK: Yes it will. And I think that makes sense to me. A conditional can be added to make the log more pedantically correct. But this path feels sufficient for expressing the state. At best, I would maybe add a comment in the code. But not change the log itself.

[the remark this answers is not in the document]
trag1c: How do you feel about dropping the log entirely? I think it's confusing (especially for first time runners) and at the same time I think it's too annoying to do nice conditionals for such a minor thing 😄
<sub>+ I wasn't really in favor of adding it to begin with lol</sub>

PR #321, opened 2025-08-27
refactor: replace pyright -> basedpyright

### LINKED ISSUE #317: Replace pyright with basedpyright
I suggest using <https://github.com/DetachHead/basedpyright> over `pyright`. It is basically pyright, but better. With bug fixes and pylance features. To migrate, you would basically only need to replace all instances of pyright with basedpyright. Migration is super minimal. No new typing errors are exposed with the use of basedpyright.

Is this a desirable change? Please let me know either way. And I can migrate to `basedpyright` if wanted.

CoderJoshDK:
Closes #317 

If you decide that you do want to swap this out, you can use this PR. If not, feel free to just close both issue and PR.

But I also made this because A) it was super fast to do. And B) you can check out the branch to try it out yourself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,6 @@ jobs:
       - name: pytest
         if: ${{ !cancelled() && steps.sync.conclusion == 'success' }}
         run: uv run pytest -vv
-      - name: pyright
+      - name: basedpyright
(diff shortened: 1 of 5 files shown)


00-kat (approved):
Approve pending discussion in #317.

PR #314, opened 2025-08-25
Introduce Loguru for logging

00-kat:
Closes #313.

As per [Loguru's documentation on security considerations](https://loguru.readthedocs.io/en/stable/resources/recipes.html#:~:text=LOGURU_DIAGNOSE), `LOGURU_DIAGNOSE=NO` should almost certainly be set in the production environment itself (which likely needs an issue in meta; let me know if I should create one).

Edit: I just realized, this is PR #100π!

PR #322, opened 2025-08-27
fix: ensure secrets are not leaked in logs

CoderJoshDK:
A few changes are implemented here:
* Pydantic uses [SecretStr](https://docs.pydantic.dev/2.0/usage/types/secrets/) (Helping prevent secret from making it out during config construction)
  * Modify usage of changed type to extract string value
* Setup #314 Loguru _before_ constructing discord bot, to fully capture all of its messages

A side effect is that the init logs now look like:
```log
2025-08-27 14:06:43.058 | WARNING  | discord.client:__init__:318 - PyNaCl is not installed, voice will NOT be supported
2025-08-27 14:06:43.173 | INFO     | sentry_sdk.integrations.logging:sentry_patched_callhandlers:128 - logging in using static token
2025-08-27 14:06:44.120 | INFO     | sentry_sdk.integrations.logging:sentry_patched_callhandlers:128 - Shard ID None has connected to Gateway (Session ID: 18ad6340f47c8a12dd8d4de59f777768).
```
With the voice warning being populated at the start.

### DIFF
--- a/app/setup.py
+++ b/app/setup.py
@@ -11,7 +11,7 @@ import discord as dc
 from discord.ext import commands
 from githubkit import GitHub
 from loguru import logger
-from pydantic import field_validator
+from pydantic import SecretStr, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 _SENSITIVE_KEYS = (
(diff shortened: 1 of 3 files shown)


CoderJoshDK:
> does this mean RedactedStderr is no longer necessary?

The config object itself will not leak the secret. However, some consumer of the string could still do it. The redaction is an extra safeguard. IMO as long as there is no noticeable performance issue, may as well keep it in for now. And just mental note that you can drop it if it starts giving you too much headache

00-kat (approved):

app/setup.py
```diff
@@ -109,7 +98,9 @@ def __getattr__(self, attr: str) -> Any:
     def write(self, content: str) -> None:
         for key in _SENSITIVE_KEYS:
             if secret := getattr(config, key):
```
00-kat, line 100: I don't really have an opinion on whether it is desirable, but I wonder if this should be something like:

```suggestion
for key, value in config:
    if isinstance(key, SecretStr):
```

so as to avoid duplicating the secret strings separately.

00-kat:
> With the voice warning being populated at the start.

Also: I'm not sure what's changed; the PyNaCL log has been there since b22688b1. Is it the `sentry_patched_callhandlers` that's new? (I don't have Sentry enabled locally so I just get `discord.client:login:643`/etc.)

Issue #313, opened 2025-08-24
Introduce `loguru` for proper logging


trag1c:
It was pointed out in https://github.com/ghostty-org/discord-bot/pull/312#discussion_r2296503548 that stdout prints are probably not the best way to do logging 😅 Using a proper logger interface will also let us throw in additional debug logs (opt-in behind a flag) in the future to help with development.

The scope of this task is limited to replacing all prints with [`loguru`](https://github.com/Delgan/loguru) calls, the debug log part should be done separately.

PR #347, opened 2025-09-10
Add assorted log messages

00-kat:
Follow-up to #313.

The following are done:
- Everything in utils.
- Everything in common except `message_moving.py`.
- Every file (not directory) under `app`.
- `components/developer.py` and `components/accept_invite.py`.

These are just some of the easier ones, I'm going to continue with the other components but I just wanted to open a PR of what I have right now.

Log levels and messages open to bikeshedding, of course.

### DIFF
--- a/app/common/linker.py
+++ b/app/common/linker.py
@@ -7,9 +7,10 @@ from types import MappingProxyType
 from typing import TYPE_CHECKING, ClassVar, Self, final
 
 import discord as dc
+from loguru import logger
 
 from app.errors import SafeView
-from app.utils import is_dm, is_mod, safe_edit
+from app.utils import is_dm, is_mod, pretty_print_account, safe_edit
 
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable
@@ -24,8 +25,10 @@ class ProcessedMessage:
 
 
 async def remove_view_after_delay(message: dc.Message, delay: float = 30.0) -> None:
+    logger.trace("waiting {}s to remove view of {}", delay, message)
     await asyncio.sleep(delay)
     with safe_edit:
+        logger.debug("removing view of {}", message)
         await message.edit(view=None)
 
 
@@ -44,9 +47,11 @@ class MessageLinker:
         return dt.datetime.now(tz=dt.UTC) - dt.timedelta(hours=24)
 
     def freeze(self, message: dc.Message) -> None:
+        logger.debug("freezing message {}", message)
         self._frozen.add(message)
 
     def unfreeze(self, message: dc.Message) -> None:
+        logger.debug("unfreezing message {}", message)
         self._frozen.discard(message)
 
     def is_frozen(self, message: dc.Message) -> bool:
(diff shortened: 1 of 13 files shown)


app/components/developer.py
```diff
@@ -40,4 +43,5 @@ async def status(self, interaction: dc.Interaction) -> None:
 
 
 async def setup(bot: GhosttyBot) -> None:
+    logger.debug("adding Developer cog")
```
CoderJoshDK, line 46: Because of the new `bot.load_extension` function, this line should be redundant. 
```suggestion
```

app/utils/__init__.py
```diff
@@ -73,6 +73,7 @@ def is_helper(member: dc.Member) -> bool:
 
 async def try_dm(account: Account, content: str, **extras: Any) -> None:
     if account.bot:
+        logger.warning("attempted to DM a bot")
```
CoderJoshDK, line 76: You should port the pretty print logic from my other PR into here. Make the log more informative too with the account info.

app/components/accept_invite.py
```diff
@@ -30,11 +31,13 @@ async def accept_invite(self, interaction: dc.Interaction) -> None:
...
 
+        logger.debug("@{} accepted the invite", interaction.user.name)
         await self.bot.log_channel.send(
             f"{interaction.user.mention} accepted the invite!",
             allowed_mentions=dc.AllowedMentions.none(),
         )
 
 
 async def setup(bot: GhosttyBot) -> None:
+    logger.debug("adding AcceptInvite cog")
```
CoderJoshDK:
```suggestion
```
See other comment about this now being redundent.

app/components/accept_invite.py
```diff
@@ -30,11 +31,13 @@ async def accept_invite(self, interaction: dc.Interaction) -> None:
         )
         await interaction.response.send_message("Check your DMs!", ephemeral=True)
 
+        logger.debug("@{} accepted the invite", interaction.user.name)
```
CoderJoshDK, line 34: +1 on pretty print account.

app/components/developer.py
00-kat: When I wrote this I hadn't known that there was any strong connection between cogs and extensions and added both because we had an extension that added multiple cogs. You've changed it in #366 and I'm probably going to deal with the GitHub integration last (i.e.: probably after that PR is merged) so I'll remove it :).

app/common/linker.py
```diff
@@ -171,7 +195,11 @@ def __init__(self, message: dc.Message, item_count: int) -> None:
     async def _reject_early(self, interaction: dc.Interaction, action: str) -> bool:
         assert not is_dm(interaction.user)
         if interaction.user.id == self.message.author.id or is_mod(interaction.user):
+            logger.trace("{} run by author or mod", action)
             return False
+        logger.debug(
+            "{} run by {} who is not the author or a mod", action, interaction.user
```
CoderJoshDK, line 201: +1 on pretty print account

CoderJoshDK (commented):
On another pass, I think it would be a bit more informative to have the `accounts` important info show up in the logs. 

See 
https://github.com/CoderJoshDK/discord-bot/blob/094bfa6518d997da03c9fe54afd787c0c407ebfd/app/utils/__init__.py#L167-L168

app/components/accept_invite.py
00-kat: I just removed this alongside the other comment lol.

CoderJoshDK (approved):

PR #354, opened 2025-09-11
Show a commit link for the commit in `/status`

00-kat:
Screenshot of new /status output.

### DIFF
--- a/app/bot.py
+++ b/app/bot.py
@@ -62,6 +62,8 @@ class GhosttyBot(commands.Bot):
 
     @override
     async def setup_hook(self) -> None:
+        await self.bot_status.load_git_data()
+
         coros = (
             self.load_extension(f"app.components.{file.stem}")
             for file in (Path(__file__).parent / "components").iterdir()
--- a/app/components/github_integration/commits.py
+++ b/app/components/github_integration/commits.py
@@ -4,7 +4,8 @@ import asyncio
 import copy
 import re
 import string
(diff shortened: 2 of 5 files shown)


00-kat:
cc @CoderJoshDK: I can't request a review from you (thanks GitHub 🙄), but would you mind taking a look at this PR?

00-kat:
Or maybe it was that the extra space above the `Commit` header made it look less like it was hugging the username:
Screenshot in normal mode.
Or that it looks trash in compact mode:
Screenshot in compact mode.

I could fix both by adding an extra newline:
Screenshot in compact mode.
But I don't know if the amount of space added above it in normal mode is desirable:
Screenshot in normal mode.
Nor do I know if the code is considered too hacky:
```py
@property
def commit(self) -> str:
    if self.commit_data:
        # Add an extra line above the commit data to add a little bit of extra space
        # between the username and the commit, and to also make it look less weird
        # in compact mode. Since Discord strips leading and trailing whitespace, an
        # italicized space character is used instead (which works because of
        # peculiarities of Discord's Markdown parser).
        return f"_ _\n{self.commit_data}"
    commit = (
        f"[`{self._commit_hash}`](<{self.commit_url}>)"
        if self._commit_hash and self.commit_url
        else "Unknown"
    )
    return f"### Commit\n{commit}"
```

trag1c (approved):
Nice work!

Issue #350, opened 2025-09-10
Refactor `common/hooks.py`


CoderJoshDK:
Issue is written assuming #340 is merged

The usage pattern of hooks is one where a factory (`create_edit_hook`) builds the listener logic of `on_message_edit`:
https://github.com/ghostty-org/discord-bot/blob/e0f4421b43efb1ec204f50632f247b0623e99f7d/app/components/github_integration/comments/integration.py#L117-L124

However, this is funky looking logic. Besides the fact that it is highly repetitive. This section of code is very ripe for some type of refactor. Perhaps could be improved by making an abstract cog class that you plug in your various controlling functions. 

Talks about this were made on the discord + https://github.com/ghostty-org/discord-bot/pull/340#discussion_r2332174145

trag1c:
Is this issue still valid? The code has changed a bit since: https://github.com/ghostty-org/discord-bot/blob/b67b7e2dac695e23f04e37abb8ceb0e64f706612/app/components/github_integration/comments/integration.py#L114-L122

PR #340, opened 2025-09-07
refactor: components as cogs and classes

CoderJoshDK:
This PR closes #338 

There is still more room for refactoring to be done. Mainly GitHub integration component(s)!

> [!WARNING]
> This is a draft. More details will be added later! Creating PR for early exposure

PR #359, opened 2025-09-13
Move create_edit_hook() to a method of MessageLinker

00-kat:
Followup to #340. Basically the bare minimum of #350 and definitely not anywhere near what the issue actually entails.

### DIFF
--- a/app/common/linker.py
+++ b/app/common/linker.py
@@ -98,6 +98,63 @@ class MessageLinker:
                 await reply.delete()
         self.unfreeze(message)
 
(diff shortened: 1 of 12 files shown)


00-kat:
cc @CoderJoshDK: I can't request a review from you (thanks GitHub 🙄), but would you mind taking a look at this PR?

CoderJoshDK (approved):
This is a really good intermediary step! At least the code looks less funky. 
As you said, still room to be even better. Later

PR #361, opened 2025-09-13
Exit early in on_message handlers if message filter fails

### LINKED ISSUE #349: Message filter can delete a message that another hook tries to reply to (race condition)
The message filter check has to run before [`on_message` listeners] executes, and [`on_message`] should only execute if the message filter didn't delete the message, as was done with an early return previously https://github.com/ghostty-org/discord-bot/blob/2ed923b8678145098c19612666889c975f4d64ab/app/core.py#L89-L91

Otherwise, sending `#1234` in `#showcase` results in an `discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body` because it's replying to a deleted message (and if for some reason this runs before the message is deleted, there's a brief flash of Ghostty Bot sending a message only for it to be deleted again by the delete hook and the message filter, which might also result in one of them throwing a not found exception).

_Originally posted by @00-kat in https://github.com/ghostty-org/discord-bot/pull/340#discussion_r2336077689_
            

Some extra guard will need to be added to susceptible `on_message` listeners.
This is not the most common edge case but should still be addressed soon!

This issue will become relevant once #340 merges

00-kat:
Closes #349.

### DIFF
--- a/app/bot.py
+++ b/app/bot.py
@@ -124,6 +124,12 @@ class GhosttyBot(commands.Bot):
         member = self.ghostty_guild.get_member(user.id)
         return member is not None and is_mod(member)
 
+    def fails_message_filters(self, message: dc.Message) -> bool:
+        # This can't be the MessageFilter cog type because that would cause an import
+        # cycle.
+        message_filter: Any = self.get_cog("MessageFilter")
+        return message_filter and message_filter.check(message)
+
     @override
     async def on_message(self, message: dc.Message, /) -> None:
         # Ignore our own messages
@@ -133,6 +139,10 @@ class GhosttyBot(commands.Bot):
         # Simple test
         if message.guild is None and message.content == "ping":
             await try_dm(message.author, "pong")
+            return
+
+        # Don't continue if the message would be deleted by a message filter.
+        if self.fails_message_filters(message):
             return
 
         await self.process_commands(message)
--- a/app/components/github_integration/code_links.py
+++ b/app/components/github_integration/code_links.py
@@ -169,7 +169,7 @@ class CodeLinks(commands.Cog):
 
     @commands.Cog.listener("on_message")
     async def reply_with_code(self, message: dc.Message) -> None:
-        if message.author.bot:
+        if message.author.bot or self.bot.fails_message_filters(message):
             return
         output = await self.process(message)
         if output.item_count != 0:
--- a/app/components/github_integration/comments/integration.py
+++ b/app/components/github_integration/comments/integration.py
@@ -78,7 +78,7 @@ class Comments(commands.Cog):
 
     @commands.Cog.listener("on_message")
     async def reply_with_comments(self, message: dc.Message) -> None:
-        if message.author.bot:
+        if message.author.bot or self.bot.fails_message_filters(message):
             return
         embeds = [
             self.comment_to_embed(comment)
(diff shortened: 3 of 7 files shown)


app/components/message_filter.py
```diff
@@ -30,66 +30,70 @@
 _REGULAR_MESSAGE_TYPES = frozenset({dc.MessageType.default, dc.MessageType.reply})
 
 
-class _MessageFilterTuple(NamedTuple):
+class MessageFilterTuple(NamedTuple):
```
00-kat, line 33: Exported because it's in the public interface now.

app/components/message_filter.py
```diff
@@ -30,66 +30,70 @@
...
+        return None
 
-            notification = _MESSAGE_DELETION_TEMPLATE.format(
-                message.channel.mention, *msg_filter.template_fillers
-            )
-            if content_size := len(message.content):
-                notification += _MESSAGE_CONTENT_NOTICE
-            await try_dm(message.author, notification, silent=content_size > 0)
+    @commands.Cog.listener("on_message")
+    async def sweep(self, message: dc.Message) -> None:
```
00-kat, line 74: Re https://github.com/ghostty-org/discord-bot/pull/360#discussion_r2346745776: I'll change this to `async def on_message` when rebasing.

app/bot.py
```diff
@@ -124,6 +124,12 @@ def is_ghostty_mod(self, user: Account) -> bool:
         member = self.ghostty_guild.get_member(user.id)
         return member is not None and is_mod(member)
 
+    def check_message_filters(self, message: dc.Message) -> bool:
+        # This can't be a cast to the MessageFilter cog type because that would
+        # cause an import cycle.
+        message_filter = cast("Any", self.get_cog("MessageFilter"))
```
00-kat, line 130: I have no idea how else to do this; I wanted to place it in GhosttyBot to avoid half the files needing to `get_cog()` and handle the case of the cog not being loaded.

app/bot.py
```diff
@@ -135,4 +141,8 @@ async def on_message(self, message: dc.Message, /) -> None:
             await try_dm(message.author, "pong")
             return
 
+        # Don't continue if the message would be deleted by a message filter.
+        if not self.check_message_filters(message):
```
00-kat, line 145: If you're wondering why I `not` both at call sites and inside the function, it's to match the semantics of the return value of `MessageFilter.check()` itself, s.t. someone oblivious to the implementation of `check_message_filters()` would be able to just assume they're identical when you just want to check if it's truthy..

app/bot.py
trag1c:
```suggestion
        message_filter: Any = self.get_cog("MessageFilter")
```

app/bot.py
trag1c: I personally got REALLY confused when trying to understand all the flips lmao. How about renaming this to `fails_message_filters` and dropping the `not` from both the call site and inside the function? I find that way easier to read, especially in the components:
```diff
-if message.author.bot or not self.bot.check_message_filters(message):
+if message.author.bot or self.bot.fails_message_filters(message):
     return
```

CoderJoshDK (approved):
I can't think of anything better. But as we already figured out, I am not very "cleaver". Nor am I clever.

Anyways, looks good.

PR #366, opened 2025-09-15
feat: add commands for (un)loading extensions

### LINKED ISSUE #352: Loading and unloading extension logic
In the future, we should have a way to unload, load, and reload extensions. 
On that same vein, we should have some CLI entrance to only load relevant extensions; useful for testing.

### LINKED ISSUE #353: Split github extension loading
Currently (assuming #340 is merged), all GitHub components are loaded together. But really, they should be independently loaded.

This will also make doing work on #352 much easier.

### LINKED ISSUE #368: refactor Monalisten client
Currently, the Monalisten client is created - nested in the github_integration webhook logic:

https://github.com/ghostty-org/discord-bot/blob/67c130dca0898257244786c41da87f54a2a77b83/app/components/github_integration/webhooks/core.py#L37-L42

This client should not be created and passed around globally. It should instead be initialized and setup elsewhere. There are a few issues with the current setup.

> [!NOTE]
> PR should wait until monalisten 0.3.x comes out and is integrated in this repo.

## Objectives

* Import should only go down the hierarchy, not back up
  * client is defined in `./webhook/core.py` but is imported into `../commits.py`
* Try to figure out a way to unload hooks when reloading an extension that uses the client. Related #352
* All valid extensions should be tied to a single, valid cog (remove `github_integration/__init__.py:setup`)

## Potential solution(s)

Pass the client into the initialization of the bot, as an argument → setup hooks - from a helper function - during `on_ready` → setup bot reference to the listen background task → consume client only from reference to `bot`

To do this, the client could be setup inside `config.py` or it can be done inside `__main__.py`. I feel like we should remove global setup where possible. `gh` gets away from it because it was a trade-off during the large refactor. 
<details><summary>gh init</summary>
<p>

https://github.com/ghostty-org/discord-bot/blob/67c130dca0898257244786c41da87f54a2a77b83/app/config.py#L67

</p>
</details> 

The question then becomes what to do about internal event handler hooks.
<details><summary>Potential setup function example</summary>
<p>

```py
def setup(client: Monalisten) -> None:
    @client.on_internal("error")
    async def _(
        event_data: dict[str, Any],
        message: str,
        pydantic_errors: list[ErrorDetails] | None,
    ) -> None:
        msg = f"{message}\n{event_data}\n{pydantic_errors or []}"
        raise RuntimeError(msg)

    @client.on_internal("auth_issue")
    async def _(issue: AuthIssue, event_data: dict[str, Any]) -> None:
        guid = event_data.get("x-github-delivery", "<missing-guid>")
        logger.warning("token {} in event {}: {}", issue.value, guid, event_data)

    @client.on_internal("ready")
    async def _() -> None:
        logger.info("monalisten client ready")
```

</p>
</details> 

With this, who should call this setup function? The `__main__.py` file that passes in a constructed client? Or should this be done inside the bot logic itself? 

I don't think there is a perfect clean solution. Maybe something would be easier to do if you could create a class that inherits the `Monalisten` class and overrides the internal event handlers. However, these are not functions but instead a typed dict of events → list of callable. 

---

Another route you can go, is to construct many clients. One for each user (cog). That way, when the cog is unloaded, you don't have to do anything extra for ensuring that hooks are not pointing to removed functions. The setup logic would still need to be done for handling internal events (errors). But this can be abstracted to some other function.

<details><summary>Example setup function</summary>
<p>

```py
def setup() -> Monalisten:
    monalisten_client = Monalisten(
        config.github_webhook_url.get_secret_value(),
        token=config.github_webhook_secret.get_secret_value()
        if config.github_webhook_secret
        else None,
    )

    @monalisten_client.on_internal("error")
    async def _(
        event_data: dict[str, Any],
        message: str,
        pydantic_errors: list[ErrorDetails] | None,
    ) -> None:
        msg = f"{message}\n{event_data}\n{pydantic_errors or []}"
        raise RuntimeError(msg)

    @monalisten_client.on_internal("auth_issue")
    async def _(issue: AuthIssue, event_data: dict[str, Any]) -> None:
        guid = event_data.get("x-github-delivery", "<missing-guid>")
        logger.warning("token {} in event {}: {}", issue.value, guid, event_data)

    @monalisten_client.on_internal("ready")
    async def _() -> None:
        logger.info("monalisten client ready")

    return monalisten_client
```

</p>
</details> 

On cog load, you add the client's listener to the background task list (would need to be a dict now) and on unload, you remove it. I really like this solution. 

The main drawback to this solution is that you will be creating multiple consumers of the same source. I don't know what type of strain that puts on the webhook. I know it is possible since multiple bots have been running at the same time, responding to webhook events (I had mine running while trag1c was testing and both got the full event with no issue).

CoderJoshDK:
Closes #353; closes #352; closes #368

This PR loads GitHub integration components as their own cogs and introduces the commands for loading cogs (extensions).

These commands allow for hot swapping code without restarting the bot. There are exceptions to this, not all code is swapped out. So if you are changing a util / common code component, you will need to restart the whole bot. 

### reload

Unloads and loads an extension that is already loaded.
image

### unload 

Unloads an extension that is already loaded.

### load

Loads an extension that is not currently enabled. The autocomplete can't know the name of the cog before loading it, so the extension's path is provided / exposed here.
image


00-kat:
…I am not sure why GitHub says I removed a request for trag1c but shows that a review is awaited from trag1c in the sidebar 🥴; thanks GitHub.

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+                    f"{extension} is an invalid / unknown extension.", ephemeral=True
+                )
+                return
+            extensions = [extension]
+        else:
+            extensions = self.bot.get_component_extension_names()
+        reloaded_cogs: list[str] = []
+
+        await interaction.response.defer(thinking=True, ephemeral=True)
+        for cog in extensions:
```
00-kat, line 84: …and this should be called `extension` instead of `cog`.

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+    async def unload(self, interaction: dc.Interaction, extension: str) -> None:
+        # The client-side check with `default_permissions` isn't guaranteed to work.
+        if not self.bot.is_ghostty_mod(interaction.user):
+            await interaction.response.send_message(
+                "Only mods can use this command.", ephemeral=True
+            )
+            return
+        if not self.bot.is_valid_extension(extension):
+            await interaction.response.send_message(
+                f"{extension} is an invalid / unknown extension.", ephemeral=True
```
00-kat, line 104: Ditto re unknown message.

```suggestion
                f"{extension} is an invalid/unknown extension.", ephemeral=True
```

```suggestion
                f"Extension `{extension}` does not exist or is invalid.", ephemeral=True
```

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+    async def load(self, interaction: dc.Interaction, extension: str) -> None:
+        # The client-side check with `default_permissions` isn't guaranteed to work.
+        if not self.bot.is_ghostty_mod(interaction.user):
+            await interaction.response.send_message(
+                "Only mods can use this command.", ephemeral=True
+            )
+            return
+        if not self.bot.is_valid_extension(extension):
+            await interaction.response.send_message(
+                f"{extension} is an invalid / unknown extension.", ephemeral=True
```
00-kat, line 124: Ditto re unknown message.

```suggestion
                f"{extension} is an invalid/unknown extension.", ephemeral=True
```

```suggestion
                f"Extension `{extension}` does not exist or is invalid.", ephemeral=True
```

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+            )
+            return
+
+        await self.bot.unload_extension(extension)
```
00-kat, line 108: Unloading an unloaded extension errors: `discord.app_commands.errors.CommandInvokeError: Command 'unload' raised an exception: ExtensionNotLoaded: Extension 'app.components.close_help_post' has not been loaded.`

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+    @dc.app_commands.default_permissions(ban_members=True)
+    @dc.app_commands.autocomplete(extension=existing_extension_autocomplete)
+    async def unload(self, interaction: dc.Interaction, extension: str) -> None:
+        # The client-side check with `default_permissions` isn't guaranteed to work.
+        if not self.bot.is_ghostty_mod(interaction.user):
+            await interaction.response.send_message(
+                "Only mods can use this command.", ephemeral=True
+            )
+            return
+        if not self.bot.is_valid_extension(extension):
```
00-kat, line 102: For all three: it would be nice to append `app.components.` and try again if it fails with just the bare word.

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+            await interaction.response.send_message(
+                f"{extension} is an invalid / unknown extension.", ephemeral=True
+            )
+            return
+
+        await self.bot.load_extension(extension)
```
00-kat, line 128: Loading a loaded extension errors: `discord.app_commands.errors.CommandInvokeError: Command 'load' raised an exception: ExtensionAlreadyLoaded: Extension 'app.components.zig_codeblocks' is already loaded.`.

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+        await interaction.response.send_message(f"Loaded {extension}", ephemeral=True)
+
+    @load.autocomplete("extension")
+    async def unloaded_extensions_autocomplete(
```
00-kat, line 132: When starting the bot without anything special, the autocomplete shows “`app.components.github_integration`” and nothing else, but selecting that option throws `discord.app_commands.errors.CommandInvokeError: Command 'load' raised an exception: ExtensionAlreadyLoaded: Extension 'app.components.github_integration' is already loaded.`

app/components/developer.py
```diff
@@ -34,10 +48,104 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+        else:
+            extensions = self.bot.get_component_extension_names()
+        reloaded_cogs: list[str] = []
+
+        await interaction.response.defer(thinking=True, ephemeral=True)
+        for cog in extensions:
+            await self.bot.unload_extension(cog)
+            await self.bot.load_extension(cog)
+            reloaded_cogs.append(cog)
+        await interaction.followup.send(f"Reloaded {reloaded_cogs}", ephemeral=True)
```
00-kat, line 88:
```suggestion
        await interaction.followup.send(f"Reloaded `{reloaded_cogs}`", ephemeral=True)
```

Ditto every other case of an extension name being in a string shown to the user.

00-kat (changes requested):

app/components/developer.py
```diff
@@ -16,6 +18,18 @@ class Developer(commands.Cog):
     def __init__(self, bot: GhosttyBot) -> None:
         self.bot = bot
 
+    async def existing_extension_autocomplete(
+        self, _: dc.Interaction, current: str
+    ) -> list[Choice[str]]:
+        return [
+            Choice(name=name, value=cog_module.__name__)
+            for name, cog in (c for c in self.bot.cogs.items())
```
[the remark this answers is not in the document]
CoderJoshDK: Discord.py has cog classes and stores them in a `cogs` dictionary. But the function to interact with them is `extensions` ,,, so there is inconstancy somewhere, no matter what. This was my best attempt at keeping some form of consistency. But open to thoughts on how to improve the naming here. Generally, though, `cogs` == `extensions`

app/components/developer.py
00-kat: Ah right, I missed that part. I'm impartial as to whether a special-case is necessary since at least now it doesn't cause an exception to be thrown.

app/utils/__init__.py
```diff
@@ -143,3 +143,7 @@ def format_diff_note(additions: int, deletions: int, changed_files: int) -> str
     if not (changed_files and (additions or deletions)):
         return None  # Diff size unavailable
     return f"diff size: `+{additions}` `-{deletions}` ({changed_files} files changed)"
+
+
+def pretty_print_account(user: Account) -> str:
```
00-kat, line 148: Would you mind adding a test somewhat like:
```py
@given(st.text(), st.integers())
def test_pretty_print_account(name: str, id_: int) -> None:
    fake_account = cast("Account", SimpleNamespace(name=name, id=id_))
    output = pretty_print_account(fake_account)
    assert name in output
    assert str(id_) in output
```

app/bot.py
```diff
@@ -136,3 +124,42 @@ async def on_message(self, message: dc.Message, /) -> None:
             return
 
         await self.process_commands(message)
+
+    def get_component_extension_names(self) -> frozenset[str]:
+        modules: set[str] = set()
+        for module_info in pkgutil.walk_packages(
+            [Path(__file__).parent / "components"], "app.components."
+        ):
+            if not inspect.isfunction(
```
00-kat, line 133: By the way does it *have* to be a function? Would `callable()` not suffice?

00-kat (approved):

app/bot.py
CoderJoshDK: You are right, this totally would suffice.

app/components/developer.py
CoderJoshDK: Correct. The `unload_extensions` function doesn't take a `Cog`. I can make it possible for unload. But consistency says otherwise

00-kat:
> A refactor of the global client `monalisten` needs to be done to remove its background setup from a cogless setup function.

Could you create an issue to track this by the way?

00-kat (approved):

CoderJoshDK:
Converted to draft because of edge case with reloading cogs that consume monalisten. We need a way to replace the hooks associated with the cog. See #368 for discussion.

app/components/github_integration/webhooks/integration.py
```diff
@@ -0,0 +1,70 @@
...
+from app.errors import handle_error, handle_task_error
+
+if TYPE_CHECKING:
+    from monalisten import AuthIssue, Error
+
+    from app.bot import GhosttyBot
+
+
+@final
+class GitHubWebhooks(commands.Cog):
```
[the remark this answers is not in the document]
CoderJoshDK: Yea, I was copying the pattern for the files that had (have) register hooks functions. But since I removed all the cogs from those files, it would make sense to move this class below the function.

app/components/github_integration/__init__.py
```diff
@@ -1,32 +1 @@
...
-        bot.add_cog(Commits(bot, monalisten_client)),
-        bot.add_cog(CodeLinks(bot)),
-        bot.add_cog(Mentions(bot)),
-        bot.add_cog(Comments(bot)),
-    )
-
-    # Creating a strong reference
-    monalisten_task = asyncio.create_task(monalisten_client.listen())
-    monalisten_task.add_done_callback(handle_task_error)
-    bot.background_tasks.add(monalisten_task)
```
00-kat, line 32: `bot.background_tasks` isn't used anywhere now, it should be removed from `GhosttyBot`.

app/components/github_integration/models.py
```diff
@@ -149,3 +152,64 @@ def _truncate_body(cls, value: object) -> str:
...
+    message: str
+    additions: int
+    deletions: int
+    files_changed: int
+    url: str
+    date: dt.datetime | None
+    signed: bool
+
+
+class CommitCache:
```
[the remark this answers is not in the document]
00-kat: I think that's fine, but I'm not sure what @trag1c thinks.

00-kat (approved):

app/components/developer.py
```diff
@@ -39,10 +56,185 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+        await interaction.response.defer(thinking=True, ephemeral=True)
+        for ext in extensions:
+            try:
+                try:
+                    await self.bot.unload_extension(ext)
+                except commands.ExtensionNotLoaded:
+                    # If already not loaded, load it
+                    await self.bot.load_extension(ext)
+                else:
+                    await self.bot.load_extension(ext)
```
trag1c, line 103: How do you feel about this?
```suggestion
                with suppress(commands.ExtensionNotLoaded):
                    await self.bot.unload_extension(ext)
                await self.bot.load_extension(ext)
```

app/components/developer.py
```diff
@@ -39,10 +56,185 @@ async def status(self, interaction: dc.Interaction) -> None:
...
+        elif reloaded_extensions:
+            reload_message = "Reloaded:\n* " + "\n* ".join(
+                f"`{e}`" for e in reloaded_extensions
+            )
+        if failed_reloaded_extensions:
+            reload_message += "\nFailed to reload:\n* " + "\n* ".join(
+                f"`{e}`" for e in failed_reloaded_extensions
+            )
+        # Remove the newline if all extensions failed to reload
+        reload_message.strip()
```
trag1c, line 131:
```suggestion
        reload_message = reload_message.strip()
```

app/components/developer.py
```diff
@@ -17,6 +19,21 @@ class Developer(commands.Cog):
     def __init__(self, bot: GhosttyBot) -> None:
         self.bot = bot
 
+    async def existing_extension_autocomplete(
+        self, _: dc.Interaction, current: str
+    ) -> list[Choice[str]]:
+        return sorted(
+            (
+                Choice(name=name, value=cog_module.__name__)
+                for name, cog in (c for c in self.bot.cogs.items())
```
trag1c, line 28: Correct me if I'm wrong, but the genexpr feels unnecessary, since it doesn't do any copies and is still a view of the original dict:
```suggestion
                for name, cog in self.bot.cogs.items()
```

CoderJoshDK: Oh yea, this is an artifact from when this autocomplete was different. Good catch

trag1c (approved):
Thank you! And sorry for the wait.

PR #370, opened 2025-09-18
feat: autocomplete solved config options

CoderJoshDK:
It is far easier to get the already loaded sitemap data to populate the autocomplete, than to memorize the config names or go to the site for the name. It is a low cost feature. 

Tested with docs cog not loaded too. You can also type whatever you want, so the command's logic to catch invalid options is still valid. Additionally, don't have to worry about desync between site and autocomplete, since that desync would cause an error anyway. IE, reference something that is on site but not `sitemap`.

### DIFF
--- a/app/components/close_help_post.py
+++ b/app/components/close_help_post.py
@@ -90,6 +90,18 @@ class Close(commands.GroupCog, group_name="close"):
             title_prefix = additional_reply = None
         await self.close_post(interaction, "solved", title_prefix, additional_reply)
 
+    @solved.autocomplete("config_option")
+    async def option_autocomplete(
+        self, _: dc.Interaction, current: str
+    ) -> list[app_commands.Choice[str]]:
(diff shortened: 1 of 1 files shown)


app/components/close_help_post.py
```diff
@@ -90,6 +90,17 @@ async def solved(
             title_prefix = additional_reply = None
         await self.close_post(interaction, "solved", title_prefix, additional_reply)
 
+    @solved.autocomplete("config_option")
+    async def option_autocomplete(
+        self, _: dc.Interaction, current: str
+    ) -> list[app_commands.Choice[str]]:
+        docs = cast("Docs", self.bot.cogs.get("Docs", SimpleNamespace(sitemap={})))
```
00-kat, line 97: Can we not just do:

```suggestion
        if not (docs := cast("Docs | None", self.bot.cogs.get("Docs"))):
            return []
```

It reads nicer than the `SimpleNamespace` in my opinion.

00-kat (approved):

PR #383, opened 2025-09-27
feat: add more discussion state emojis

### LINKED ISSUE #380: Update discussion emojis
Discussions have more unused states. Duplicate and outdated. 

They should be supported in the embed logic for discussion. 

To take it a step further, this should also be supported in regular entity mentions. The webhook has access to `state_reason`, so it can use that. The entity logic would need to add a bit of extra support to figure out the state.

CoderJoshDK:
Closes #380

Added:

* discussion_duplicate
* discussion_outdated

Supported in webhooks and entities

### DIFF
--- a/app/components/github_integration/webhooks/discussions.py
+++ b/app/components/github_integration/webhooks/discussions.py
@@ -25,16 +25,23 @@ class DiscussionLike(Protocol):
     category: DiscussionPropCategory
     answer_html_url: str | None
     state: Literal["open", "closed", "locked", "converting", "transferring"]
+    state_reason: Literal["resolved", "outdated", "duplicate", "reopened"] | None
+
+
+def get_discussion_emoji(discussion: DiscussionLike) -> EmojiName:
+    if discussion.state_reason == "outdated":
+        return "discussion_outdated"
+    if discussion.state_reason == "duplicate":
+        return "discussion_duplicate"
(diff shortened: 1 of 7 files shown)


app/components/github_integration/webhooks/discussions.py
```diff
@@ -25,16 +25,22 @@ class DiscussionLike(Protocol):
     category: DiscussionPropCategory
     answer_html_url: str | None
     state: Literal["open", "closed", "locked", "converting", "transferring"]
+    state_reason: Literal["resolved", "outdated", "duplicate", "reopened"] | None
+
+
+def get_discussion_emoji(discussion: DiscussionLike) -> EmojiName:
+    emoji = "discussion"
+    if discussion.state_reason in ("outdated", "duplicate"):
+        emoji += "_" + discussion.state_reason
```
00-kat, line 34:
```suggestion
        emoji += f"_{discussion.state_reason}"
```

00-kat (approved):

app/components/github_integration/webhooks/discussions.py
CoderJoshDK: pyright doesn't like this change. It transforms `emoji` into a `LiteralString` and then complains since that isn't the return signature. 
But by keeping it as is, pyright is happy. And i would rather not disable pyright in the return line here.

00-kat (approved):

PR #384, opened 2025-09-28
feat: show diff from edits

### LINKED ISSUE #381: Improve PR/Issue updates
We can try to use `difflib` to improve the edited actions.

CoderJoshDK:
Closes #381

We use difflib here to show the changes between edits

In the diff, the titles are provided. Meaning, if a title changes, that is still being captured. 

image
image
image


app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
+        issue, old_issue_changes = event.issue, event.changes
 
         if issue.created_at > dt.datetime.now(tz=dt.UTC) - dt.timedelta(minutes=15):
             return
 
-        update_notes: list[str] = []
-        if changes.title:
-            update_notes.append(
-                f'Renamed from "{changes.title.from_}" to "{issue.title}"'
+        if old_issue_changes.body and old_issue_changes.body.from_:
```
CoderJoshDK, line 136: Need to check for `from_` because otherwise a title change provides a blank body

Image

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
+            # Escape code block to avoid breaking diff
+            from_file = old_issue_changes.body.from_.replace("`", r"\`").splitlines(
```
00-kat, line 138: The backslashes confused me when seeing an embed yesterday and I had reached the conclusion that you had an escaped code block in your comment for some reason.

I'd dealt with a similar situation in the past with a hacky solution which I think works better because it looks like a backquote unless you look closely: https://github.com/ghostty-org/discord-bot/blob/aa37d194c50235d14edef6d5159304c0e9f5f747/app/components/move_message.py#L892-L897

In my opinion you should pull that out into a separate function (to avoid duplication of the HACK comment), add a property test to ensure the output doesn't contain any backquotes, and use that here too.

00-kat, line 138: Single backquotes don't break the diff, it would be nice if those didn't also have backslashes before them.

```suggestion
            from_file = old_issue_changes.body.from_.replace("```", r"\`\`\`").splitlines(
```

(Note: the comment above will make this unnecessary.)

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
-
-        match update_notes:
-            case [note]:
-                content = note
-            case [note1, note2]:
-                content = f"* {note1}\n* {note2}"
-            case _:
-                return
+            to_file = (
+                issue.body.replace("`", r"\`").splitlines(keepends=True)
```
00-kat, line 142: Ditto both of the above comments.

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
 
     @webhook.event.issues.edited
     async def _(event: events.IssuesEdited) -> None:
-        issue, changes = event.issue, event.changes
+        issue, old_issue_changes = event.issue, event.changes
```
00-kat, line 131: Why? There are no other issue changes so why is this one *old* issue changes? And the other interpretation of “changes belonging to old issue” doesn't work either because there *is* no old issue.


```suggestion
        issue, changes = event.issue, event.changes
```

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
+                    from_file,
+                    to_file,
+                    fromfile=old_issue_changes.title.from_
+                    if old_issue_changes.title
+                    else issue.title,
+                    tofile=issue.title,
+                    tofiledate=issue.created_at.isoformat(),
+                )
+            )
+            available_size = 500 - (len("```diff\n\n```"))
```
00-kat, line 157:
```suggestion
            available_size = 500 - len("```diff\n\n```")
```

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
+            diff = truncate(diff, available_size)
```
00-kat, line 158:
```suggestion
            diff = truncate(diff, 500 - (len("```diff\n\n```")))
```

app/components/github_integration/webhooks/issues.py
```diff
@@ -126,26 +128,41 @@ async def _(event: events.IssuesReopened) -> None:
...
+            content = f"```diff\n{diff}\n```"
+        elif old_issue_changes.title:
+            content = (
+                f'* Renamed from "{old_issue_changes.title.from_}" to "{issue.title}"'
```
00-kat, line 162: There's only one possible note now, and the single note case didn't have a bullet before: https://github.com/ghostty-org/discord-bot/blob/aa37d194c50235d14edef6d5159304c0e9f5f747/app/components/github_integration/webhooks/issues.py#L143-L144

```suggestion
                f'Renamed from "{old_issue_changes.title.from_}" to "{issue.title}"'
```

app/components/github_integration/webhooks/prs.py
```diff
@@ -97,24 +99,37 @@ async def _(event: events.PullRequestReopened) -> None:
 
     @webhook.event.pull_request.edited
```
00-kat, line 100: Could the issue and PR edited handlers be abstracted into a helper function? There's *much* more duplication now compared to the previous functions.

00-kat (changes requested):

app/components/github_integration/webhooks/issues.py
CoderJoshDK: I tried the triple, but if they somehow send \`\`\`\`\` it still breaks. Now, to be fair, this is so highly unlickely. But technically possible, and why I didn't go with it in the first place. 

`````py
print("it does still render properly on GH")
`````

File: app/components/github_integration/webhooks/core.py [hunk 9]
```diff
@@ -20,6 +23,7 @@
     from app.config import WebhookFeedType
 
 type EmbedColor = Literal["green", "red", "purple", "gray", "orange", "blue"]
+CODEBLOCK = re.compile(r"`{3,}")
```
00-kat, line 26:
```suggestion
CODE_BLOCK = re.compile(r"`{3,}")

type EmbedColor = Literal["green", "red", "purple", "gray", "orange", "blue"]
```

app/components/github_integration/webhooks/core.py
```diff
@@ -86,6 +90,67 @@ def dict(self, bot: GhosttyBot) -> dict[str, str | None]:
...
+        event.issue if isinstance(event, events.IssuesEdited) else event.pull_request
+    )
+
+    if event_object.created_at > dt.datetime.now(tz=dt.UTC) - dt.timedelta(minutes=15):
+        return
+
+    changes = event.changes
+    if changes.body and changes.body.from_:
+        # HACK: replace all 3+ backticks with reverse primes to avoid breaking the diff
+        # block, while maintaining the intent.
```
00-kat, line 119: Commas are cool but a pause there reads a bit weird (it feels like the last phrase is hanging off).

```suggestion
        # block while maintaining the intent.
```

(A comma after `reverse primes` doesn't sound odd to me in my opinion, so you could add one there if you wanted to do so.)

00-kat (approved):

PR #389, opened 2025-09-29
feat: auto cleanup webhooks

### LINKED ISSUE #387: Handle discussion issues and html body better
If the top of an issue doesn't have the moved discussion header, 
Image

The code that cleans it up, doesn't. 
https://github.com/ghostty-org/discord-bot/blob/aa37d194c50235d14edef6d5159304c0e9f5f747/app/components/github_integration/webhooks/issues.py#L30-L38

This is a good opportunity to generally improve the display of content. More specifically, cleanup raw HTML 
Image

<details><summary>Bike shed</summary>
<p>

Convert links and references in the same manner that GH does. For example, `#387` becomes a link #387 (This has a lot of edge cases and might be impossible. But it would also inherently solve the converted discussion issue)

This would require a bit of movie magic to pull off. And so maybe should only be done against full fat GitHub links. 
One edge case to be really careful about is truncated text. We don't want to truncate a created link. This can be difficult because you would need to know the character length of the shown text and not the markdown, and take that into account. 

</p>
</details>

CoderJoshDK:
Closes #387 

This PR cleans up embeds sent from webhook logic. This handles the `<sup>` and `<sub>` tags and shortens long GH links into hyperlinks, when possible and logical. This **does not** convert things like `#123` into a hyperlink.

Created issues might also have `div` tags that should be removed. A tag that comes from converting a discussion into an issue. It just takes up space that we can ignore on creation.

### DIFF
--- a/app/components/github_integration/webhooks/issues.py
+++ b/app/components/github_integration/webhooks/issues.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import re
-from functools import partial
 from typing import TYPE_CHECKING, Any, Protocol, cast
 
 from loguru import logger
@@ -15,51 +14,18 @@ from app.components.github_integration.webhooks.utils import (
 
 if TYPE_CHECKING:
     from githubkit.typing import Missing
-    from githubkit.versions.latest.models import RepositoryWebhooks
     from monalisten import Monalisten, events
 
     from app.bot import EmojiName, GhosttyBot
 
 
-GITHUB_DISCUSSION_URL = re.compile(
-    # Ignore if already inside a link block
-    r"(?<!\()"
-        r"https://github\.com/"
-        r"(?P<owner>\b[a-zA-Z0-9\-]+/)"
-        r"(?P<repo>\b[a-zA-Z0-9\-\._]+)"
-        r"(?P<sep>/(?:issues|pull|discussions)/)"
-        r"(?P<number>\d+)"
-    r"(?!\))"
-)  # fmt: skip
-SUP_HTML = re.compile(r"\s*<sup>((?:.|\s)+?)</sup>\s*")
 DISUCSSION_DIV_TAG = re.compile(
     r"\s*<div type='discussions-op-text'>((?:.|\s)*?)\s*</div>\s*", re.MULTILINE
 )
 
 
-def shorten_same_repo_links(
-    origin_repo: RepositoryWebhooks, matchobj: re.Match[str]
-) -> str:
-    if (
-        matchobj.group("owner") == origin_repo.owner.name
-        and matchobj.group("repo") == origin_repo.name
-    ):
-        # Only short hand if link comes from same repo
-        return f"[#{matchobj.group('number')}]({matchobj.group()})"
-    return matchobj.group()
-
-
-def reformat_converted_discussion_header(
-    body: str | None, repo: RepositoryWebhooks
-) -> str | None:
-    if not body:
-        return body
-    body = SUP_HTML.sub(
-        lambda x: "".join(f"\n-# {line}\n" for line in x.group(1).splitlines()), body
-    )
-    body = DISUCSSION_DIV_TAG.sub(r"\g<1>", body)
-    body = GITHUB_DISCUSSION_URL.sub(partial(shorten_same_repo_links, repo), body)
-    return body  # noqa: RET504
+def remove_discussion_div(body: str | None) -> str | None:
+    return DISUCSSION_DIV_TAG.sub(r"\g<1>", body) if body else body
 
 
 class IssueLike(Protocol):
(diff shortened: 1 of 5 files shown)


app/components/github_integration/webhooks/issues.py
```diff
@@ -15,27 +16,50 @@
...
-CONVERTED_DISCUSSION_HEADER = re.compile(
-    r"\s*### Discussed in https://github.com/.*?/discussions/(?P<discussion_number>\d+)"
-    r"\s*<div type='discussions-op-text'>"
-    r"\s*<sup>(?P<subtext>.+?)</sup>",
-    re.MULTILINE,
+
+GITHUB_DISCUSSION_URL = re.compile(
+    # Ignore if already inside a link block
+    r"(?<!\()"
+        r"https://github.com/"
```
00-kat, line 28:
```suggestion
        r"https?://(?:www\.)?github\.com/"
```

xref #210.

app/components/github_integration/webhooks/utils.py
```diff
@@ -135,10 +159,19 @@ async def send_embed(  # noqa: PLR0913
     *,
     color: EmbedColor | None = None,
     feed_type: WebhookFeedType = "main",
+    origin_repo: RepositoryWebhooks | None = None,
 ) -> None:
+    content_dict = content.dict
```
trag1c, line 164: This feels unnecessary.

app/components/github_integration/webhooks/utils.py
```diff
@@ -12,12 +13,23 @@
...
 
 if TYPE_CHECKING:
-    from githubkit.versions.latest.models import SimpleUser
+    from githubkit.versions.latest.models import RepositoryWebhooks, SimpleUser
 
     from app.bot import EmojiName, GhosttyBot
     from app.config import WebhookFeedType
 
 CODEBLOCK = re.compile(r"`{3,}")
+SUP_HTML = re.compile(r"\s*<su[pb]>(.+?)</su[pb]>\s*")
```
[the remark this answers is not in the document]
CoderJoshDK: We can either make `SUP_HTML` and `SUB_HTML` their own regex variables. Or ,,, just accept that this broken html will result in broken output. I avoided the double variable for performance reasons. But this code path will be hit very infrequent, and the cost is low. So I can make them two variables.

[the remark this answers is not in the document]
trag1c: We could use a backreference (like `\s*<(su[pb])>(.+?)</\1>\s*`), but that introduces a group we'd have to discard later in processing 🤔 (since you can't have backreferences to non-capturing groups)

app/components/github_integration/webhooks/utils.py
```diff
@@ -127,6 +139,18 @@ async def send_edit_difference(
     )
 
 
+def _shorten_same_repo_links(
+    origin_repo: RepositoryWebhooks, matchobj: re.Match[str]
+) -> str:
+    if (
+        matchobj.group("owner") == origin_repo.owner.name
+        and matchobj.group("repo") == origin_repo.name
+    ):
```
trag1c, line 148: This is never true, because `matchobj.group("owner")` includes a trailing slash, so `"ghostty-org/" == "ghostty-org"` fails. A simple change of `+/)` to `+)/` at the end of L27 fixes this.

trag1c (changes requested):

app/components/github_integration/webhooks/utils.py
```diff
@@ -135,7 +159,15 @@ async def send_embed(  # noqa: PLR0913
...
     color: EmbedColor | None = None,
     feed_type: WebhookFeedType = "main",
+    origin_repo: RepositoryWebhooks | None = None,
 ) -> None:
+    if origin_repo and (body := content.dict.get("description")):
+        body = SUBTEXT_HTML.sub(r"\n-# \g<2>\n", body)
+        body = GITHUB_DISCUSSION_URL.sub(
+            partial(_shorten_same_repo_links, origin_repo), body
+        )
+        content.dict["description"] = body
```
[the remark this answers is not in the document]
CoderJoshDK: Ah, yea. I knew there was a reason why I pulled the dict out earlier. This works too.

app/components/github_integration/webhooks/utils.py
```diff
@@ -127,6 +140,18 @@ async def send_edit_difference(
     )
 
 
+def _shorten_same_repo_links(
+    origin_repo: RepositoryWebhooks, matchobj: re.Match[str]
+) -> str:
+    if (
+        matchobj.group("owner") == origin_repo.owner.name
+        and matchobj.group("repo") == origin_repo.name
+    ):
```
trag1c, line 149: This is still problematic :( GitHub can just omit setting `origin_repo.owner.name`, so I think we should use `.full_name` which is always present:
```suggestion
    owner, _, repo = origin_repo.full_name.partition("/")
    if matchobj["owner"] == owner and matchobj["repo"] == repo:
```

app/components/github_integration/webhooks/issues.py
```diff
@@ -18,23 +18,14 @@
...
-    r"\s*<div type='discussions-op-text'>"
-    r"\s*<sup>(?P<subtext>.+?)</sup>",
-    re.MULTILINE,
-)
-
 
-def reformat_converted_discussion_header(body: str | None, repo_url: str) -> str | None:
-    if body is None or not (match := CONVERTED_DISCUSSION_HEADER.match(body)):
-        return body
+DISUCSSION_DIV_TAG = re.compile(
```
00-kat, line 22:
```suggestion
DISCUSSION_DIV_TAG = re.compile(
```

00-kat (approved):

00-kat (approved):

PR #398, opened 2025-10-04
fix: get none from cache when fetch fails

### LINKED ISSUE #396: GraphQLFailed: data={'repository': None} errors=[GraphQLError(type='NOT_FOUND', message="Could not resolve to a Repository with the name 'master/INTRODUCTION.md'.", locations=[SourceLocation(line=3, column=3)], path=['repository'], extensions=None)] extensions=None
Sentry Issue: [DISCORD-BOT-42](https://ghostty.sentry.io/issues/6922707879/?referrer=github_integration)

```
RequestFailed: Response(404 Not Found, data_model=<class 'githubkit.versions.v2022_11_28.models.group_0006.BasicError'>)
  File "app/components/github_integration/entities/cache.py", line 26, in fetch
    entity = (await self.gh.rest.issues.async_get(*key)).parsed_data

GraphQLFailed: data={'repository': None} errors=[GraphQLError(type='NOT_FOUND', message="Could not resolve to a Repository with the name 'master/INTRODUCTION.md'.", locations=[SourceLocation(line=3, column=3)], path=['repository'], extensions=None)] extensions=None
(1 additional frame(s) were not displayed)
...
  File "app/common/cache.py", line 39, in get
    await self._refresh(key)
  File "app/common/cache.py", line 31, in _refresh
    await self.fetch(key)
  File "app/components/github_integration/entities/cache.py", line 33, in fetch
    self[key] = await get_discussion(*key)
  File "app/components/github_integration/entities/discussions.py", line 34, in get_discussion
    resp = await gh.graphql.arequest(
```

CoderJoshDK:
Fixes #396 

A fetch can fail if the thing that is being fetched doesn't exist. It is impossible to know if something exists before checking. For example, `foo/bar#123` is valid looking, but doesn't point to anything.

To keep the overall design, `get` needs to return `None` when the fetch fails. And all instances of `fetch` should handle their own errors. We don't want to suppress errors inside the cache as to keep logic separate. And let the caller decide if an error can be recovered from.

This does mean that `fetch` should ensure to suppress exceptions when settings some value, letting the rest of the `cache.get` play out and return `None`.


CoderJoshDK:
xkcd does set the cache on a failed fetch. So why not do that for the other operations? 
XKCD is very unlikely to be triggered in a false positive case. While the other options are more likely.

00-kat (approved):

app/components/github_integration/entities/cache.py
```diff
@@ -30,7 +31,8 @@ async def fetch(self, key: EntitySignature) -> None:
                 model = PullRequest
             self[key] = model.model_validate(entity, from_attributes=True)
         except RequestFailed:
-            self[key] = await get_discussion(*key)
+            with suppress(GraphQLFailed):
+                self[key] = await get_discussion(*key)
```
00-kat, line 35: Should the `GraphQLFailed` suppress be done in `get_discussion()` (which would return `None` if it failed)? That would force callers to handle it via the return type instead of it being easy to miss as it is now.

app/components/github_integration/entities/discussions.py
```diff
@@ -30,10 +34,12 @@
...
 
-async def get_discussion(org: str, name: str, number: int) -> Discussion:
-    resp = await gh.graphql.arequest(
-        DISCUSSION_QUERY, variables={"number": number, "org": org, "repo": name}
-    )
-    data = resp["repository"]["discussion"]
-    data["answered_by"] = (answer := data.pop("answer")) and answer["user"]
-    return Discussion(**data)
+async def get_discussion(org: str, name: str, number: int) -> Discussion | None:
+    with suppress(GraphQLFailed):
```
00-kat, line 38: I think this should just be a try-catch instead of a suppress + return in both branches.

00-kat (approved):

app/components/github_integration/comments/discussions.py
```diff
@@ -64,15 +66,22 @@ def _encode_discussion_comment_id(comment_id: int) -> str:
...
-    discussion["answered_by"] = (answer := discussion.pop("answer")) and answer["user"]
-    return Comment(
-        **resp["node"],
-        entity_gist=entity_gist,
-        entity=Discussion(**discussion),
-    )
+async def get_discussion_comment(
+    entity_gist: EntityGist, comment_id: int
+) -> Comment | None:
+    try:
```
00-kat, line 72: Actually I think the try-catch should be scoped to only the `gh.graphql.arequest` call.

app/components/github_integration/entities/discussions.py
```diff
@@ -30,10 +32,13 @@
...
 
-async def get_discussion(org: str, name: str, number: int) -> Discussion:
-    resp = await gh.graphql.arequest(
-        DISCUSSION_QUERY, variables={"number": number, "org": org, "repo": name}
-    )
-    data = resp["repository"]["discussion"]
-    data["answered_by"] = (answer := data.pop("answer")) and answer["user"]
-    return Discussion(**data)
+async def get_discussion(org: str, name: str, number: int) -> Discussion | None:
+    try:
```
00-kat, line 36: Ditto.

00-kat (changes requested):

trag1c (approved):

PR #407, opened 2025-10-15
refactor: remove comment's global reference of cache & gh

CoderJoshDK:
Only removes the global state from comments fetching.


CoderJoshDK:
This code is going to be refactored anyways. Closing this example now. Can reference to it later if needed.

PR #409, opened 2025-10-16
Handle interactive xkcds

00-kat:
List of all of them: https://www.explainxkcd.com/wiki/index.php/Category:Interactive_comics.

### DIFF
--- a/app/components/xkcd_mentions.py
+++ b/app/components/xkcd_mentions.py
@@ -34,6 +34,7 @@ class XKCD(BaseModel):
     year: int
     title: str
     img: str
+    link: str
     transcript: str
     alt: str
     extra_parts: dict[str, str] | None = None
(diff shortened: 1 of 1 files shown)


app/components/xkcd_mentions.py
```diff
@@ -89,11 +93,32 @@ def get_embed(xkcd: XKCDResult) -> dc.Embed:
...
+                # Some interactive comics have https://imgs.xkcd.com/comics/ as
+                # their image, which results in no image showing because that
+                # URL is not an image and also 403s. Check the extension
+                # instead of hardcoding that URL since there could be other
+                # comics with a different problematic image URL.
+                _, _, ext = xkcd.img.rpartition(".")
+                if f".{ext}" in SUPPORTED_IMAGE_FORMATS:
+                    embed.set_image(url=xkcd.img)
+                elif xkcd.transcript:
+                    embed.description = xkcd.transcript
```
CoderJoshDK: Is this ever false? Should this just be an `else`?

00-kat: Yeah newer xkcds don't have transcripts. E.g. the current one: https://xkcd.com/3155/info.0.json.

PR #411, opened 2025-10-16
chore: remove bsky embed replacement

CoderJoshDK:
As brought up by @oshdubh, the default bluesky embed is actually really good. There is no need to replace it since they do a killer job.

But if that is not reason enough, the default endpoint does a better job with some embeds:

image


00-kat:
I *was* going to bring this up during CR but the main thing that bsky's embeds don't include in reactions so I decided why not, lol.

PR #412, opened 2025-10-17
Move regular message types set to utils and on_message precondition to helper function

00-kat:
See individual commit descriptions.

### DIFF
--- a/app/bot.py
+++ b/app/bot.py
@@ -17,7 +17,7 @@ from loguru import logger
 
 from app.errors import handle_error, interaction_error_handler
 from app.status import BotStatus
-from app.utils import is_mod, pretty_print_account, try_dm
+from app.utils import REGULAR_MESSAGE_TYPES, is_mod, pretty_print_account, try_dm
 
 if TYPE_CHECKING:
     from githubkit import GitHub, TokenAuthStrategy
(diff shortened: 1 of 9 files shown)


app/utils/__init__.py
```diff
@@ -43,6 +43,17 @@
 _INVITE_LINK_REGEX = re.compile(r"\b(?:https?://)?(discord\.gg/[^\s]+)\b")
 _ORDERED_LIST_REGEX = re.compile(r"^(\d+)\. (.*)")
 
+# Regular types taken from the description of
+# https://discordpy.readthedocs.io/en/stable/api.html#discord.Message.system_content.
+REGULAR_MESSAGE_TYPES = frozenset({
+    dc.MessageType.default,
+    dc.MessageType.reply,
+})
+BOT_COMMAND_MESSAGE_TYPES = frozenset({
```
00-kat: Not sure if I should keep this (`BOT_COMMAND_MESSAGE_TYPES`) in message_moving until it's required elsewhere.

app/bot.py
```diff
@@ -177,6 +177,15 @@ def fails_message_filters(self, message: dc.Message) -> bool:
         message_filter: Any = self.get_cog("MessageFilter")
         return message_filter and message_filter.check(message)
 
+    def on_message_preconditions_fail(
+        self, message: dc.Message, *, ignore_system_messages: bool = True
```
00-kat: Also not sure if `ignore_system_messages` is necessary since I don't think there's any harm to always checking (it's also currently unused).

app/utils/__init__.py
trag1c: I think it's fine to keep in utils.

PR #416, opened 2025-10-19
feat: custom message_filter_passed dispatch

### LINKED ISSUE #415: Custom dispatch for message filter
Right now, every `on_message` has to do a check for if it is going to be filtered out due to the message filter. But that burden should be centralized into one location. (Or one logical path). The dispatch function can be used to create a custom event that passes this filter.

This idea/request comes from @onerandomusername, thank you very much for the suggestion!

CoderJoshDK:
Closes #415 

Instead of requiring ever `on_message` to check some common `on_message_preconditions_fail`, we can provide our own `on_message` that already has this check completed. Dispatch a custom event `on_message_filter_passed`. IE, after messages are already filtered. 

This is done inside the bot file and not the message filter cog, because we don't want to rely on the cog existing for our other cogs to trigger. 

This PR is also a performance improvement (by a tiny amount) because we don't have to recalculate if a message passes the filter every single time. 

## Other minor changes

1. Removed `await self.process_commands(message)` because we also removed command prefixes. So this line no longer should ever trigger anything. 
2. Linker edit ignores bot messages. This check used to happen on the `on_message` path. So it was moved over into the linker itself. It also optionally happened implicitly behind the `interactor`. So either way, logic remains as it did before. 
3. In tests, `spawn_message` now expects you to explicitly state if it is a bot or a user. This is to play nice with the change mentioned right before this.

### DIFF
--- a/app/bot.py
+++ b/app/bot.py
@@ -189,7 +189,7 @@ class GhosttyBot(commands.Bot):
             return
 
         if not self._fails_message_filters(message):
-            self.dispatch("post_message_filter", message)
+            self.dispatch("message_filter_passed", message)
 
     @classmethod
     def get_component_extension_names(cls) -> frozenset[str]:
--- a/app/components/developer.py
+++ b/app/components/developer.py
@@ -32,7 +32,7 @@ class Developer(commands.Cog):
             ),
         )
 
-    @commands.Cog.listener("on_post_message_filter")
+    @commands.Cog.listener("on_message_filter_passed")
     async def sync_handler(self, message: dc.Message) -> None:
         # Handle !sync command. This can't be a slash command because this command is
         # the one that actually adds the slash commands in the first place. This does
--- a/app/components/fixup_embeds.py
+++ b/app/components/fixup_embeds.py
@@ -88,7 +88,7 @@ class FixupEmbeds(commands.Cog):
         )
 
     @commands.Cog.listener()
-    async def on_post_message_filter(self, message: dc.Message) -> None:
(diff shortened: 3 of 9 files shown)


app/bot.py
```diff
@@ -181,7 +181,7 @@ def on_message_preconditions_fail(self, message: dc.Message) -> bool:
         return (
             message.author.bot
```
CoderJoshDK: Technically, this boolean is evaluated twice in the path now. However, it is super cheap and I think being explicit is fine here.

app/bot.py
```diff
@@ -196,11 +196,8 @@ async def on_message(self, message: dc.Message, /) -> None:
             await try_dm(message.author, "pong")
             return
 
-        # Don't continue if the message would be deleted by a message filter.
-        if self.fails_message_filters(message):
-            return
-
-        await self.process_commands(message)
+        if not self.on_message_preconditions_fail(message):
+            self.dispatch("post_message_filter", message)
```
00-kat: I feel like this should be called `message_filter_pass` or `…passed` or `…passing`, since the handler is named `on_{name}`.

app/bot.py
CoderJoshDK: In my memory, the check was to ignore messages from bots. But on second pass, it was ignoring itself. So, actually ignore me.

app/bot.py
CoderJoshDK: This is much better, too. Now I can grep `on_message` to find all `on_message` instances, regardless of if special dispatch.

00-kat (approved):

trag1c (approved):

PR #417, opened 2025-10-19
feat: bump diff size truncate size

### LINKED ISSUE #408: Consider bumping diff's max size when a title change is present
Re https://github.com/ghostty-org/discord-bot/pull/406#discussion_r2430697281, https://discord.com/channels/1281624934636064811/1281624935558807678/1426546704815231091, https://discord.com/channels/1281624934636064811/1281624935558807678/1427267265740542085, and https://discord.com/channels/1281624934636064811/1281624935558807678/1428210366617227366.

Essentially the issue is that the title eats up a *lot* of characters, and we only have 500.

Or maybe it should just be bumped regardless of whether a title change is present, since diffs have double the content (previous + current).

CoderJoshDK:
Closes #408

Provides a path to explicitly set the embed description. This allows us to avoid the 500 character truncate limit.

### DIFF
--- a/app/components/github_integration/webhooks/utils.py
+++ b/app/components/github_integration/webhooks/utils.py
@@ -61,9 +61,9 @@ class EmbedContent(NamedTuple):
     def dict(self) -> EmbedContentArgs:
         args: EmbedContentArgs = {"title": self.title, "url": self.url}
         if self.description:
-            # If provided a description explicitly, don't truncate. However, discord has
-            # a description character size limit.
-            args["description"] = truncate(self.description, 6000)
+            # If a description is provided explicitly, don't truncate. However, Discord
+            # has a description character size limit.
+            args["description"] = truncate(self.description, 4096)
         elif self.body:
             args["description"] = truncate(self.body, 500)
         return args


app/components/github_integration/webhooks/utils.py
```diff
@@ -55,12 +55,17 @@ class EmbedContent(NamedTuple):
...
 
     @property
     def dict(self) -> EmbedContentArgs:
         args: EmbedContentArgs = {"title": self.title, "url": self.url}
         if self.body:
             args["description"] = truncate(self.body, 500)
+        if self.description:
+            # If provided a description explicitly, don't truncate. However, discord has
+            # a description character size limit.
+            args["description"] = truncate(self.description, 6000)
```
00-kat: The diff is horrible but I essentially moved the description if above the body if and made the body if an elif.

```suggestion
        if self.description:
            # If provided a description explicitly, don't truncate. However, discord has
            # a description character size limit.
            args["description"] = truncate(self.description, 6000)
        elif self.body:
            args["description"] = truncate(self.body, 500)
```

00-kat (changes requested):

app/components/github_integration/webhooks/utils.py
```diff
@@ -55,11 +55,16 @@ class EmbedContent(NamedTuple):
...
     body: str | None = None
+    description: str | None = None
 
     @property
     def dict(self) -> EmbedContentArgs:
         args: EmbedContentArgs = {"title": self.title, "url": self.url}
-        if self.body:
+        if self.description:
+            # If provided a description explicitly, don't truncate. However, discord has
+            # a description character size limit.
```
00-kat:
```suggestion
            # If a description is provided explicitly, don't truncate. However, discord
            # has a description character size limit.
```

trag1c (approved):

PR #421, opened 2025-10-20
chore: update dependencies

CoderJoshDK:
Move `msgpack-types` as a dev dependency. It is only needed for development and not for runtime logic. 

Update general dependencies. Mainly capture the latest sentry-sdk patch. While at it, update the CI too.


trag1c (approved):
Thanks!

PR #422, opened 2025-10-22
Greatly improve Reddit subdomain handling

00-kat:


CoderJoshDK (commented):
This should really be made into a library ... 

but otherwise, looks good. Style nits. But logic looks good.

PR #428, opened 2025-12-11
chore(deps): update dependencies

CoderJoshDK:
Can't update githubkit to latest version due to conflict with monalisten

Also updated checkout to v6

pytest now disables the terminfo by default. So I added the flag to enable it in the just file.


00-kat (approved):
(Approve pending discussion on `-p terminalprogress`, but merging as-is would be fine too.)

trag1c (approved):

PR #441, opened 2026-01-17
Remove config and gh globals

00-kat:
xref #466.

There are probably ways to make this better, but I can't think of anything without increasing complexity by an order of magnitude, so I decided to just PR my current solution.

Points of discussion:
  - Is this useful? I originally did this because GhosttyBot was given a config and GitHub client but most users use globals instead, but this would help with isolating core GitHub functionality from the rest of the bot (as was considered recently on Discord) in the future. I decided to PR it before asking because it's hard to know without seeing what the code might look like.
  - Would a better form of dependency injection other than passing multiple arguments everywhere be desirable?
  - (*see PR comments below*)

### DIFF
--- a/app/components/github_integration/comments/discussions.py
+++ b/app/components/github_integration/comments/discussions.py
@@ -4,11 +4,11 @@ from typing import TYPE_CHECKING
 import msgpack
 from githubkit.exception import GraphQLFailed
 
-from app.components.github_integration.entities.cache import entity_cache
 from app.components.github_integration.models import Comment, Discussion
 
 if TYPE_CHECKING:
     from app.components.github_integration.models import EntityGist
+    from toolbox.misc import GH
 
 DISCUSSION_COMMENT_QUERY = """
 query getDiscussionComment($id: ID!) {
@@ -50,12 +50,12 @@ query getDiscussionComment($id: ID!) {
 
 
 async def get_discussion_comment(
-    entity_gist: EntityGist, comment_id: int
+    gh: GH, entity_gist: EntityGist, comment_id: int
 ) -> Comment | None:
     packed = msgpack.packb([0, 0, comment_id])
     node_id = "DC_" + urlsafe_b64encode(packed).decode()
     try:
-        resp = await entity_cache.gh.graphql.arequest(
+        resp = await gh.graphql.arequest(
             DISCUSSION_COMMENT_QUERY, variables={"id": node_id}
         )
     except GraphQLFailed:
--- a/app/components/github_integration/entities/cache.py
+++ b/app/components/github_integration/entities/cache.py
@@ -1,65 +1,30 @@
-from typing import TYPE_CHECKING, final, override
+from typing import final, override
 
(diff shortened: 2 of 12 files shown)


app/components/github_integration/comments/discussions.py
```diff
@@ -55,7 +55,7 @@ async def get_discussion_comment(
     packed = msgpack.packb([0, 0, comment_id])
     node_id = "DC_" + urlsafe_b64encode(packed).decode()
     try:
-        resp = await gh.graphql.arequest(
+        resp = await entity_cache.gh.graphql.arequest(
```
00-kat: I can't think of a cleaner way to get a GitHub client to many of these functions. The comment cache doesn't have a client, and it would be far too much weird code to give the comment cache a client just for its callers to also use the mention cache with its own client; every function directly using `entity_cache.gh` is solely called from the comment cache.

app/components/github_integration/entities/cache.py
```diff
@@ -1,23 +1,35 @@
...
+    multiple cogs register a client, it is undefined which client will be used, but
+    clients of unloaded cogs will not be used; this will not be an issue if one client
+    is used by all cogs.
+    """
+
+    def __init__(self, **ttr: float) -> None:
         super().__init__(**ttr)
-        self.gh: GH = gh
+        self._current_gh: GH | None = None
+        self._registered_ghs: dict[commands.Cog, GH] = {}
```
00-kat: I didn't consider the less hacky method of loading every cog that uses this cache because that requires a list of users here, instead of closer to the actual usage. However, the “registration” mechanic is a bit weird, so it might be worth it despite the reduction in code locality.

app/bot.py
```diff
@@ -37,9 +38,12 @@
 ]
 
 
+type Emojis = MappingProxyType[EmojiName, dc.Emoji]
```
trag1c:
```suggestion
type Emojis = Mapping[EmojiName, dc.Emoji]
```

app/components/github_integration/webhooks/utils.py
```diff
@@ -157,11 +168,15 @@ def _shorten_same_repo_links(
...
-    bot: GhosttyBot,
     actor: SimpleUser,
     content: EmbedContent,
     footer: Footer,
     *,
+    emojis: Emojis,
+    # This function is frequently partially applied, and webhook channels usually aren't
+    # available yet when partial() is called, so take a function that returns
+    # a (probably memoized) result as an argument instead.
+    channels: Callable[[], WebhookChannels],
```
CoderJoshDK: I think it is worth the effort to make this `Callable[[], WebhookChannels] | WebhookChannels`. Allowing the partial to take either. Instead of doing these pass through lambdas.

00-kat:
> If the barebones API (that bases on globals) is not convenient enough

The stdlib API seems to have fared perfectly—it's parallels the current usage, with the problem arising from them being global being resolved, so I've just used the stdlib API in #454. Once again thanks for the suggestion!

PR #446, opened 2026-01-25
Don't consider snowflakes to be a commit SHA

### LINKED ISSUE #444: Ignore Discord IDs for commit links
We're currently doing a lot of requests any time a person/channel gets mentioned. Someone trying to use an ~18-char long string of digits to refer to a Ghostty commit is just not happening (and when it is they should use `ghostty@` or `main@`).

00-kat:
Fixes #444. This does mean things like `<<061a0ae5656c05525aa812f474dee2ed32700125` no longer match, but that's an extremely unlikely case.


CoderJoshDK (approved):
Sometimes regex be looking funny. But yep, this checks out.

PR #468, opened 2026-02-21
Fix emoji access race

00-kat:
See individual commit descriptions.

The first commit is co-authored Joshie as they suggested using `asyncio.Event` a long time ago on Discord.

### DIFF
--- a/app/bot.py
+++ b/app/bot.py
@@ -39,6 +39,8 @@ EmojiName = Literal[
     "pull_open",
 ]
 
+_EMOJI_NAMES = frozenset(get_args(EmojiName))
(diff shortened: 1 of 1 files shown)


00-kat:
Marking as draft: I want to cherry-pick one more commit.

trag1c (approved):
Not fond of the `dc.Emoji | str` but otherwise this is a good solution, thanks!

00-kat:
It was always defaulted to a `"❓️"` (except the defaulting happened late), I have no idea why it type-checked before but doesn't now without the `| str`.

PR #492, opened 2026-03-15
ignore bare number SHAs for commit mentions

### LINKED ISSUE #444: Ignore Discord IDs for commit links

trag1c:
Closes #444.

The original PR missed some things (e.g. emojis or links) and it's unlikely that someone will mention a digit-only SHA anyway, so we can just ignore them.

From a quick check it looks like this will drop the number of failed commit requests by ~95%, and the number of failed requests in general by ~26%.


00-kat (approved):
Fair enough.

PR #519, opened 2026-04-11
Bump dependencies

00-kat:
Summary:
```perl
              aiohttp   v3.13.3 -> v3.13.5
                anyio   v4.12.1 -> v4.13.0
                attrs   v25.4.0 -> v26.1.0
         basedpyright   v1.38.2 -> v1.39.0
            githubkit   v0.15.0 -> v0.15.2
           hypothesis  v6.151.9 -> v6.151.12
nodejs-wheel-binaries  v24.14.0 -> v24.14.1
             pygments   v2.19.2 -> v2.20.0
               pytest    v9.0.2 -> v9.0.3
                 ruff   v0.15.5 -> v0.15.10
           sentry-sdk   v2.54.0 -> v2.57.0
   astral-sh/setup-uv    v7.3.1 -> v8.0.0
                   uv   v0.10.4 -> v0.11.6
```

Nothing particularly interesting changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
-  UV_VERSION: 0.10.4
+  UV_VERSION: 0.11.6
   UV_NO_SYNC: 1
   UV_LOCKED: 1
   PY_COLORS: 1
@@ -24,7 +24,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Setup uv
-        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098  # v7.3.1
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57  # v8.0.0
         with:
           version: ${{ env.UV_VERSION }}
       - name: Environment
@@ -62,7 +62,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Setup uv
-        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098  # v7.3.1
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57  # v8.0.0
         with:
           version: ${{ env.UV_VERSION }}
       - name: Environment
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Setup uv
-        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098  # v7.3.1
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57  # v8.0.0
```
00-kat, line 27: Apparently this is an immutable release ([source]), which means it's safe to use:

```suggestion
        uses: astral-sh/setup-uv@v8.0.0
```

Should I do that (perhaps with an `# immutable` comment)?

[source]: https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0

trag1c: That source says that the hash is still better 🤔

00-kat: I did see that too which is why I kept the hash but I couldn't discern the reason from [GitHub's docs](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) so I dunno.

00-kat: Tangent: I can't see why the hash is better but I can see one reason for why it's *worse*: nothing checks that comment, so it could be a lie (or mistake) and everyone might forget to double-check.

trag1c: I went ahead and asked. For
> nothing checks that comment

we could look into using zizmor which does that.

trag1c: Got answers from Jelle Zijlstra and Joshie :)
> I think a hash is still better for a few (fairly paranoid reasons): you don't have to trust that GH implemented immutable releases correctly, and it's easier to audit a GH Actions file for safety if all tags use a hash

> That last one feels pretty relevant. You know a tag is immutable because it’s a hash. You don’t know it’s immutable if it’s just a tag

00-kat: Ah. That's a fair take I guess.

trag1c (approved):

PR #521, opened 2026-04-11
Use `tool.ruff.lint.flake8-import-conventions.extend-aliases` instead of `aliases` to keep defaults

00-kat:
See https://docs.astral.sh/ruff/settings/#lint_flake8-import-conventions_aliases.


trag1c (approved):
We'll likely never need any of the defaults, but good to know!

PR #529, opened 2026-04-17
linker: don't process frozen messages

### LINKED ISSUE #527: Frozen messages still cause requests
Scenario:
* user sends `#1234`
* bot replies with the entity
* user freezes the message
* user edits the message to `#1234 #1235`
* bot doesn't change its message, but still makes a request to fetch `#1235`

trag1c:
Fixes #527. This sounds too easy so maybe I'm missing something here?

### DIFF
--- a/packages/toolbox/src/toolbox/linker.py
+++ b/packages/toolbox/src/toolbox/linker.py
@@ -143,6 +143,10 @@ class MessageLinker:
             self.unlink(before)
             return
 
+        if self.is_frozen(before):
+            logger.trace("skipping frozen message {msg}", msg=before)
+            return
+
         old_output = await message_processor(before)
         new_output = await message_processor(after)
         if old_output == new_output:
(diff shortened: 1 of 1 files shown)


packages/toolbox/src/toolbox/linker.py
00-kat:
```suggestion

```

00-kat (changes requested):

packages/toolbox/src/toolbox/linker.py
00-kat: Amazing suggestion diff GitHub 🫩 anyway note that this is a different one from the one you added.

00-kat:
> This sounds too easy so maybe I'm missing something here?

I think it's right, from what I can tell the freeze was carefully done after the get because there used to be code between the freeze and the get in the past: https://github.com/ghostty-org/discord-bot/commit/ec4c5d3fe6953f9b51cd9983952b50b3a17330fc#diff-d9b3a454dc21efcdbf2748b9f8808d03f9188dc12c9815fc00b5e479daedfacfR103-R108.

packages/toolbox/src/toolbox/linker.py
trag1c: Oh good catch, thanks!

00-kat (approved):

PR #536, opened 2026-05-03
Organize toolbox tests the same way as toolbox

00-kat:
Previously these utilities were all in `app/utils`, but now that it's been moved to toolbox and split up, there is no `toolbox/utils`, but rather `toolbox/{discord,github,messages,misc}`, etc.

This was a pretty easy change to make, but of course the diff is beyond useless; I found that grabbing `test_utils.py` from the previous commit then diffing that file with the new ones one by one was pretty effective.

### DIFF
--- a/packages/toolbox/tests/test_discord_utils.py
+++ b/packages/toolbox/tests/test_discord_utils.py
@@ -1,12 +1,10 @@
 import datetime as dt
-import subprocess
-import sys
 from typing import TYPE_CHECKING, Any
 from unittest.mock import Mock
 
 import discord as dc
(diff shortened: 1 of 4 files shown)


CoderJoshDK (approved):
This is ,,, fine. I don't love super tiny test files. But this does make sense to do. So I don't mind.

trag1c (approved):

00-kat:
> I don't love super tiny test files.

Same; I do make `test_github_utils.py` larger in future commits, but `test_messages_utils.py` is still smaller than I would like it.

PR #541, opened 2026-05-06
Merge review request webhook events into one

00-kat:
Closes #509.

I unfortunately wasn't able to test this *too* well, as I couldn't get GitHub to let me request a review from someone; there are many tests though to hopefully catch everything, and I made sure sending it to Discord works fine via [`Monalisten.dispatch_event`], the code for which can be found at 3170176c3f741b5270c619bb1cdf8743e92a07dd. It is probably advisable to test it once in ghrebote with real data.

This PR isn't nearly as large as it looks; 70% of it is tests, since the logic is very intricate and is devoid of API calls.

[`Monalisten.dispatch_event`]: https://github.com/trag1c/monalisten#testing-with-manual-events

### DIFF
--- a/app/components/github_integration/webhooks/review_summary.py
+++ b/app/components/github_integration/webhooks/review_summary.py
@@ -1,0 +1,272 @@
+import asyncio
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, NamedTuple, Self
+
+from loguru import logger
+from monalisten import events
+
+from app.components.github_integration.models import GitHubTeam, GitHubUser
+from toolbox.misc import drain_queue
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterable
+
+
+REVIEW_REQUEST_GROUP_TIMEOUT = 5  # seconds
+
+
+type Reviewer = GitHubUser | GitHubTeam
+type ReviewRequestsModified = (
+    events.PullRequestReviewRequested | events.PullRequestReviewRequestRemoved
+)
+type ReviewPools = dict[ReviewPoolKey, asyncio.Queue[ReviewRequestsModified]]
+
+
+class ReviewPoolKey(NamedTuple):
+    """
+    The minimum amount of data needed to differentiate semantic groups of review
+    requests or removals.
+    """
+
+    pr_number: int
+    actor_id: int
+
+
+async def handle_review_request(
+    review_pools: ReviewPools,
+    event: ReviewRequestsModified,
+    *,
+    timeout: int | None = REVIEW_REQUEST_GROUP_TIMEOUT,  # noqa: ASYNC109
+) -> ReviewRequestSummary | None:
+    key = ReviewPoolKey(pr_number=event.pull_request.number, actor_id=event.sender.id)
+
+    # If we already have a queue for this PR+user combination, simply collect to it.
+    if q := review_pools.get(key):
+        logger.info("found existing review pool for key {key}", key=key)
+        await q.put(event)
+        return None
+
+    logger.info(
+        "creating new review pool #{n} for key {key}", n=len(review_pools) + 1, key=key
+    )
+    q = asyncio.Queue[ReviewRequestsModified]()
+    review_pools[key] = q
+    # This won't throw QueueShutDown as it cannot be shut down yet: the timeout
+    # hasn't been set.
+    await q.put(event)
+    # The coroutine calling this function will see this queue through to the end;
+    # future review_request() calls for this queue will terminate immediately after
+    # updating the queue.
+
+    async def timer() -> None:
+        if timeout is None:
+            return
+        await asyncio.sleep(timeout)
+        del review_pools[key]
+        # Signal the timeout to the queue.
+        q.shutdown()
+        logger.info(
+            "shut down review pool {key} ({n} remain)", key=key, n=len(review_pools)
+        )
+
+    async with asyncio.TaskGroup() as group:  # transparently handle cancelation.
+        group.create_task(timer())
+        summary = group.create_task(ReviewRequestSummary.collect(drain_queue(q)))
+
+    return summary.result()
+
+
+def _parse_reviewer(event: ReviewRequestsModified) -> Reviewer | None:
+    # Abusing duck typing here because the API models are insufferable.
+    event_dyn: Any = event
+    if hasattr(event_dyn, "requested_team"):
+        return GitHubTeam(name=event_dyn.requested_team.name)
+    if requested_reviewer := getattr(event_dyn, "requested_reviewer", None):
+        return GitHubUser(**requested_reviewer.model_dump())
+    return None
+
+
+@dataclass(kw_only=True)
+class ReviewRequestSummary:
+    requests: set[Reviewer] = field(default_factory=set)
+    removals: set[Reviewer] = field(default_factory=set)
+    # When a review was requested from somebody, then later removed.
+    accidental_requests: set[Reviewer] = field(default_factory=set)
+    # When a review request for somebody was removed, then later re-requested.
+    rerequests: set[Reviewer] = field(default_factory=set)
+    unknown_requests: int = 0  # count.
+    unknown_removals: int = 0  # count.
+
+    def __bool__(self) -> bool:
+        return bool(
+            self.requests
+            or self.removals
+            or self.accidental_requests
+            or self.rerequests
+            or self.unknown_requests
+            or self.unknown_removals
+        )
+
+    @staticmethod
+    def _categorize(
+        r: Reviewer,
+        *,
+        default_set: set[Reviewer],
+        opposite_set: set[Reviewer],
+        null_set: set[Reviewer],
+        opposite_null_set: set[Reviewer],
+    ) -> None:
+        """
+        Categorize a reviewer given its event and the opposite event's sets. This
+        function mutates the passed sets.
+
+        Arguments:
+            default_set -- The set associated with this event (`removals`/`requests`).
+            opposite_set -- The set associated with the opposite event (`removals` for
+                requests; `requests` for removals).
+            null_set -- The set associated with events of this type that were later
+                undone; i.e.: requesting a review then removing it again (represented by
+                `accidental_requests`), or removing a request then requesting it again
+                (represented by `rerequests`).
+            opposite_null_set -- Likewise, but for the opposite event.
+
+        Note that the null sets are sets of reviewers with a null (zero *delta*, and
+        aren't always-null (empty) sets.
+        """
+        # NOTE: an invariant is that every reviewer is only present in one set at
+        # a time, which is asserted in every branch below as a precaution.
+        if r in null_set:
+            # Either:
+            #   - a review was requested (r put in `requests`), then removed (r put in
+            #     `accidental_requests`), then requested again (this event).
+            #   - a review request was removed (r put in `removals`), then requested (r
+            #     put in `rerequests`), then removed again (this event).
+            # These request/remove cycles can be arbitrarily long, so we'll only show
+            # the difference between the initial and final states—that is, we'll undo
+            # the previous update which moved it from the default set to the null set.
+            for s in (default_set, opposite_set, opposite_null_set):
+                assert r not in s
+            default_set.add(r)
+            null_set.discard(r)
+        elif r in opposite_null_set:
+            # Either:
+            #   - a review request was removed (r put in `removals`), then requested (r
+            #     put in `rerequests`), then requested again (this event).
+            #   - a review was requested (r put in `requests`), then removed (r put in
+            #     `accidental_requests`), then removed again (this event).
+            # This is impossible in theory but random chance writes better applications
+            # than GitHub, so just ignore it.
+            for s in (default_set, opposite_set):
+                assert r not in s
+            logger.warning(
+                "ignoring theoretically impossible duplicate review undo event for {r}",
+                r=r.name,
+            )
+        elif r in opposite_set:
+            # Either a review request was removed then requested again, or the opposite.
+            # This event is the second one, so we want to move it to the *opposite* null
+            # set—that of the *first*.
+            assert r not in default_set
+            opposite_null_set.add(r)
+            opposite_set.discard(r)
+        elif r in default_set:
+            # A review was requested twice in a row, or a review request was likewise
+            # removed. Again, this is impossible in theory, but GitHub's webhook event
+            # synchronization is, to say the least, nonexistent.
+            logger.warning(
+                "ignoring theoretically impossible duplicate review event for {r}",
+                r=r.name,
+            )
+        else:
+            default_set.add(r)
+
+    @classmethod
+    async def collect(cls, it: AsyncIterable[ReviewRequestsModified]) -> Self:
+        summary = cls()
+        async for ev in it:
+            r = _parse_reviewer(ev)
+            if isinstance(ev, events.PullRequestReviewRequested):
+                if r is None:
+                    summary.unknown_requests += 1
+                else:
+                    summary._categorize(
+                        r,
+                        default_set=summary.requests,
+                        opposite_set=summary.removals,
+                        null_set=summary.accidental_requests,
+                        opposite_null_set=summary.rerequests,
+                    )
+            else:  # noqa: PLR5501 # the symmetry improves clarity.
+                if r is None:
+                    summary.unknown_removals += 1
+                else:
+                    summary._categorize(
+                        r,
+                        default_set=summary.removals,
+                        opposite_set=summary.requests,
+                        null_set=summary.rerequests,
+                        opposite_null_set=summary.accidental_requests,
+                    )
+        return summary
+
+    @staticmethod
+    def _format_reviewers(
+        reviewers: set[Reviewer], unknown: int = 0, *, heading: str | None = None
+    ) -> str:
+        heading_ = f"**{heading}** " * (heading is not None)
+        match sorted(reviewers, key=lambda r: r.name):
+            case [] if unknown:
+                # Only unknowns.
+                return f"{heading_}from {unknown} unknown users"
+            case []:
+                return ""
+            # Unknowns go on a separate line, so if there is both a reviewer and
+            # unknowns, go down the last case instead so that they're formatted on
+            # multiple lines.
+            case [reviewer] if not unknown:
+                return f"{heading_}from {reviewer.format()}"
+            case rs:
+                return (
+                    f"{heading_}from:\n"
+                    + "\n".join(f"- {r.format()}" for r in rs)
+                    + (f"\n- {unknown} unknown users" if unknown else "")
+                )
+
+    def format(self) -> tuple[str, str]:
+        """Return the appropriate title and body, respectively, in a 2-tuple."""
+        have_requests = bool(self.requests or self.unknown_requests)
+        have_removals = bool(self.removals or self.unknown_removals)
+        if (
+            self.accidental_requests
+            or self.rerequests
+            or not (have_requests ^ have_removals)
+        ):
+            sections = (
+                self._format_reviewers(
+                    self.requests, self.unknown_requests, heading="Requested review"
+                ),
+                self._format_reviewers(
+                    self.removals,
+                    self.unknown_removals,
+                    heading="Removed review request",
+                ),
+                self._format_reviewers(
+                    self.accidental_requests, heading="Accidentally requested review"
+                ),
+                self._format_reviewers(
+                    self.rerequests, heading="Removed, then requested review"
+                ),
+            )
+            return "updated review requests", "\n\n".join(filter(None, sections))
+        if have_requests:
+            return "requested review", self._format_reviewers(
+                self.requests, self.unknown_requests
+            )
+        if have_removals:
+            return "removed review request", self._format_reviewers(
+                self.removals, self.unknown_removals
+            )
+        # Fail loudly in case someone messes up when updating the Boolean algebra hell
+        # above.
+        msg = "unreachable"
+        raise AssertionError(msg)
--- a/packages/toolbox/src/toolbox/misc.py
+++ b/packages/toolbox/src/toolbox/misc.py
@@ -1,16 +1,19 @@
 import asyncio
 import re
 import subprocess
+from contextlib import suppress
 from typing import TYPE_CHECKING, Any, Literal
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, AsyncIterable
+    from collections.abc import AsyncGenerator, AsyncIterable, Sequence
 
 __all__ = (
     "COLOR_PALETTE",
     "URL_REGEX",
     "aenumerate",
     "async_process_check_output",
+    "drain_queue",
+    "seq_to_aiter",
     "truncate",
 )
 
@@ -29,6 +32,17 @@ URL_REGEX = re.compile(
     r"https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b"
     r"(?:[-a-zA-Z0-9()@:%_\+.~#?&//=]*)"
 )
+
+
+async def seq_to_aiter[T](elems: Sequence[T]) -> AsyncGenerator[T]:
+    """
+    Convert any sequence to an asynchronous iterator. Synchronous iterables in general
+    aren't supported as this function doesn't magically insert await points into
+    them—computing the elements would block the scheduler until it finishes if it takes
+    a nontrivial amount of time!
+    """
+    for elem in elems:
+        yield elem
 
 
 def truncate(s: str, length: int, *, suffix: str = "…") -> str:
@@ -62,3 +76,10 @@ async def async_process_check_output(program: str, *args: str, **kwargs: Any) ->
             stderr=proc.stderr and await proc.stderr.read(),
         )
     return (await proc.stdout.read()).decode()
+
+
+async def drain_queue[T](queue: asyncio.Queue[T]) -> AsyncGenerator[T]:
+    """Yield `queue.get()` repeatedly until it has been shut down."""
+    with suppress(asyncio.QueueShutDown):
+        while True:
+            yield await queue.get()
--- a/packages/toolbox/tests/utils.py
+++ b/packages/toolbox/tests/utils.py
@@ -1,0 +1,18 @@
+from contextlib import suppress
+
+from hypothesis import strategies as st
+
+
+def any_comparables() -> st.SearchStrategy[object]:
+    """Generate anything Hypothesis can generate that equals itself."""
+
+    def is_comparable(x: object) -> bool:
+        # Suppress all exceptions because some values (such as signaling NaNs:
+        # Decimal("sNaN")) throw an exception when compared.
+        with suppress(Exception):
+            # Disable the self-comparison lint as Hypothesis might be able to generate
+            # values other than NaN that don't compare equal to themself.
+            return x == x  # noqa: PLR0124
+        return False
+
+    return st.from_type(object).filter(is_comparable)
--- a/tests/webhooks/test_review_summary.py
+++ b/tests/webhooks/test_review_summary.py
@@ -1,0 +1,701 @@
+# pyright: reportPrivateUsage=false
+
+import asyncio
+from unittest.mock import Mock
+
+import pytest
+from hypothesis import assume, given, settings
+from hypothesis import strategies as st
+from monalisten.events import PullRequestReviewRequested as PRReviewReq
+from monalisten.events import PullRequestReviewRequestRemoved as PRReviewReqRm
+
+from tests.webhooks.utils import (
+    DEFAULT_PR_SENDER_ID,
+    add_pr_metadata,
+    dummy_user,
+    github_teams,
+    github_users,
+    review_request_modification_events,
+    review_request_summaries,
+    reviewers,
+    team_review_request,
+    unknown_review_request,
+    user_review_request,
+)
+
+from app.components.github_integration.models import GitHubTeam, GitHubUser
+from app.components.github_integration.webhooks.review_summary import (
+    Reviewer,
+    ReviewPoolKey,
+    ReviewPools,
+    ReviewRequestsModified,
+    ReviewRequestSummary,
+    _parse_reviewer,
+    handle_review_request,
+)
+from toolbox.misc import seq_to_aiter
+
+
+def test_empty_review_request_summary_falsy() -> None:
+    assert not ReviewRequestSummary()
+
+
+@given(github_users())
+def test_parse_reviewer_user(user: GitHubUser) -> None:
+    event = user_review_request(
+        name=user.name, html_url=user.url, avatar_url=user.icon_url
+    )
+    assert _parse_reviewer(event) == user
+
+
+@given(github_teams())
+def test_parse_reviewer_team(team: GitHubTeam) -> None:
+    assert _parse_reviewer(team_review_request(name=team.name)) == team
+
+
+def test_parse_reviewer_none() -> None:
+    assert _parse_reviewer(Mock(())) is None
+
+
+@settings(max_examples=25)
+@given(st.lists(review_request_modification_events(include_unknowns=False)))
+async def test_collect_review_request_summary_contains_all_reviewers(
+    requests: list[ReviewRequestsModified],
+) -> None:
+    reviewers = set(map(_parse_reviewer, requests))
+    s = await ReviewRequestSummary.collect(seq_to_aiter(requests))
+    assert s.requests | s.removals | s.accidental_requests | s.rerequests == reviewers
+
+
+@settings(max_examples=25)
+@given(st.lists(review_request_modification_events()))
+async def test_collect_review_request_summary_has_no_duplicate_reviewers(
+    requests: list[ReviewRequestsModified],
+) -> None:
+    s = await ReviewRequestSummary.collect(seq_to_aiter(requests))
+    assert len(s.requests | s.removals | s.accidental_requests | s.rerequests) == (
+        len(s.requests)
+        + len(s.removals)
+        + len(s.accidental_requests)
+        + len(s.rerequests)
+    )
+
+
+REVIEW_REQUEST_COLLECTION_TESTS = [
+    (
+        [user_review_request(PRReviewReq, name="foo")],
+        ReviewRequestSummary(requests={dummy_user(name="foo")}),
+    ),
+    (
+        [team_review_request(PRReviewReq, name="ghostty-org/spook")],
+        ReviewRequestSummary(requests={GitHubTeam(name="ghostty-org/spook")}),
+    ),
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            team_review_request(PRReviewReqRm, name="bar"),
+        ],
+        ReviewRequestSummary(removals={dummy_user(name="foo"), GitHubTeam(name="bar")}),
+    ),
+    (
+        [
+            team_review_request(PRReviewReq, name="foo"),
+            user_review_request(PRReviewReqRm, name="bar"),
+        ],
+        ReviewRequestSummary(
+            requests={GitHubTeam(name="foo")}, removals={dummy_user(name="bar")}
+        ),
+    ),
+    (
+        [
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+        ],
+        ReviewRequestSummary(accidental_requests={GitHubTeam(name="foo")}),
+    ),
+    (
+        [
+            user_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+        ],
+        ReviewRequestSummary(
+            requests={dummy_user(name="foo")}, removals={GitHubTeam(name="foo")}
+        ),
+    ),
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="bar"),
+            team_review_request(PRReviewReq, name="baz"),
+        ],
+        ReviewRequestSummary(
+            requests={GitHubTeam(name="baz")},
+            removals={GitHubTeam(name="bar")},
+            rerequests={dummy_user(name="foo")},
+        ),
+    ),
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+        ],
+        ReviewRequestSummary(removals={dummy_user(name="foo")}),
+    ),
+    (
+        [
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReq, name="foo"),
+        ],
+        ReviewRequestSummary(requests={GitHubTeam(name="foo")}),
+    ),
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReq, name="foo"),
+            user_review_request(PRReviewReq, name="foo"),
+        ],
+        ReviewRequestSummary(rerequests={dummy_user(name="foo")}),
+    ),
+    (
+        [
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+        ],
+        ReviewRequestSummary(accidental_requests={GitHubTeam(name="foo")}),
+    ),
+    (
+        [
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+            team_review_request(PRReviewReq, name="foo"),
+            team_review_request(PRReviewReqRm, name="foo"),
+            team_review_request(PRReviewReq, name="foo"),
+        ],
+        ReviewRequestSummary(requests={GitHubTeam(name="foo")}),
+    ),
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReq, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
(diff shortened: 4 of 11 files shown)


tests/webhooks/utils.py
```diff
@@ -41,3 +53,128 @@ def make_pr(
...
+    )
+
+
+@st.composite
+def github_teams(draw: st.DrawFn) -> GitHubTeam:
+    return GitHubTeam(name=draw(st.text()))
+
+
+def reviewers() -> st.SearchStrategy[Reviewer]:
+    return github_teams() | github_users()
```
trag1c, line 75: I don't know about `github_users`, but can't these two be simplified to this?
```suggestion
github_teams = st.text().map(GitHubTeam)

reviewers = github_teams | github_users()
```
I think (not 100% sure) it doesn't make a difference for Hypothesis and they use this approach themselves (e.g. in Django extras).

tests/webhooks/utils.py
```diff
@@ -41,3 +53,128 @@ def make_pr(
...
+
+@st.composite
+def review_request_summaries(draw: st.DrawFn) -> ReviewRequestSummary:
+    return ReviewRequestSummary(
+        requests=draw(st.sets(reviewers())),
+        removals=draw(st.sets(reviewers())),
+        no_requests=draw(st.sets(reviewers())),
+        rerequests=draw(st.sets(reviewers())),
+        unknown_requests=draw(st.integers(0)),
+        unknown_removals=draw(st.integers(0)),
```
trag1c: Can we make the `min_value=` explicit? Or, even better, make some aliases:
```py
NonnegInts = st.integers(min_value=0)
PosInts = st.integers(min_value=1)
```

tests/webhooks/test_review_summary.py
```diff
@@ -0,0 +1,694 @@
...
+    await run_review_request_collection_test(pools, 12, requests, result)
+
+
+# Do it again, except in parallel, to catch cases where having multiple PRs in the queue
+# makes it malfunction (i.e.: bugs caused by accidental dependencies between separate
+# PRs). The other test has MUCH better error messages, so both shall be run instead of
+# only this one, despite this one being a superset of the other one.
+@settings(max_examples=10)
+@given(st.permutations(REVIEW_REQUEST_COLLECTION_TESTS))
+async def test_handle_review_request_parllel_collection(
```
trag1c: Typo:
```suggestion
async def test_handle_review_request_parallel_collection(
```

tests/webhooks/test_review_summary.py
```diff
@@ -0,0 +1,694 @@
...
+) -> None:
+    pools: ReviewPools = {}
+    await run_review_request_collection_test(pools, 12, requests, result)
+
+
+# Do it again, except in parallel, to catch cases where having multiple PRs in the queue
+# makes it malfunction (i.e.: bugs caused by accidental dependencies between separate
+# PRs). The other test has MUCH better error messages, so both shall be run instead of
+# only this one, despite this one being a superset of the other one.
+@settings(max_examples=10)
```
trag1c, line 292: What's the motivation behind limiting this to 10 examples (same goes for other tests being limited to 25 or 50)?

tests/webhooks/test_review_summary.py
```diff
@@ -0,0 +1,694 @@
...
+    requests: list[Mock], result: ReviewRequestSummary
+) -> None:
+    pools: ReviewPools = {}
+    await run_review_request_collection_test(pools, 12, requests, result)
+
+
+# Do it again, except in parallel, to catch cases where having multiple PRs in the queue
+# makes it malfunction (i.e.: bugs caused by accidental dependencies between separate
+# PRs). The other test has MUCH better error messages, so both shall be run instead of
+# only this one, despite this one being a superset of the other one.
```
trag1c: I found this comment a bit confusing because of the assumption that you read the previous test's comment first. Can we start with `Same as the test above,` or ``Same as `test_handle_review_request_collection`,`` instead?

tests/webhooks/test_review_summary.py
```diff
@@ -0,0 +1,694 @@
...
+    (
+        [
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+            user_review_request(PRReviewReqRm, name="foo"),
+        ],
+        ReviewRequestSummary(removals={dummy_user(name="foo")}),  # + a log message.
```
trag1c: Is it useful to have these comments? I'd consider dropping them or including the logs in the cases to test that as well.

tests/webhooks/test_review_summary.py
```diff
@@ -0,0 +1,694 @@
...
+    ReviewRequestsModified,
+    ReviewRequestSummary,
+    _parse_reviewer,
+    handle_review_request,
+)
+from toolbox.misc import list_to_aiter
+
+
+def test_empty_review_request_summary_falsy() -> None:
+    assert not ReviewRequestSummary()
```
trag1c, line 40: Think we could use a truthy test?

packages/toolbox/tests/utils.py
```diff
@@ -0,0 +1,25 @@
+from contextlib import suppress
+
+from hypothesis import strategies as st
+
+
+@st.composite
+def anys(draw: st.DrawFn) -> object:
+    """Generate anything Hypothesis can generate."""
+    any_type: type[type] = draw(st.from_type(type))
+    return draw(st.from_type(any_type))
```
trag1c: Is this any different from using `st.from_type(object)` as a strategy?

packages/toolbox/src/toolbox/misc.py
```diff
@@ -31,6 +34,11 @@
 )
 
 
+async def list_to_aiter[T](elems: list[T]) -> AsyncGenerator[T]:
+    for elem in elems:
+        yield elem
```
trag1c: Why did you choose not to generalize this to all iterables?

packages/toolbox/src/toolbox/misc.py
```diff
@@ -62,3 +70,10 @@ async def async_process_check_output(program: str, *args: str, **kwargs: Any) ->
             stderr=proc.stderr and await proc.stderr.read(),
         )
     return (await proc.stdout.read()).decode()
+
+
+async def drain_queue[T](queue: asyncio.Queue[T]) -> AsyncGenerator[T]:
+    """Yield queue.get() repeatedly until it has been shut down."""
```
trag1c: For consistency with some other docstrings you've added (I'm also fine with flipping this around and dropping the backticks from other places, like `tests/webhooks/utils.py:177`).
```suggestion
    """Yield `queue.get()` repeatedly until it has been shut down."""
```

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,283 @@
...
+async def handle_review_request(
+    review_pools: ReviewPools,
+    event: ReviewRequestsModified,
+    *,
+    timeout: int | None = REVIEW_REQUEST_GROUP_TIMEOUT,  # noqa: ASYNC109
+    force_shutdown: bool = False,
+) -> ReviewRequestSummary | None:
+    """
+    force_shutdown can be set to forcibly shut down the pool associated with this
+    event's pull request and return a review summary immediately, bypassing the timeout.
```
trag1c: Same as my other backtick comment.
```suggestion
    `force_shutdown` can be set to forcibly shut down the pool associated with this
    `event`'s pull request and return a review summary immediately, bypassing the
    timeout.
```

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,283 @@
...
+            await q.put(event)
+        return None
+
+    # Make a new queue and store it; the dictionary set doesn't need synchronization
+    # as the GIL makes it atomic.
+    logger.info(
+        "creating new review pool #{n} for key {key}", n=len(review_pools) + 1, key=key
+    )
+    q = asyncio.Queue[ReviewRequestsModified]()
+    review_pools[key] = q
```
trag1c: I still didn't entirely digest the logic, but what does "the dictionary set" mean here?

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,283 @@
...
+        return GitHubUser(**requested_reviewer.model_dump())
+    return None
+
+
+@dataclass(kw_only=True)
+class ReviewRequestSummary:
+    requests: set[Reviewer] = field(default_factory=set)
+    removals: set[Reviewer] = field(default_factory=set)
+    # When a review was requested from somebody, then later removed.
+    no_requests: set[Reviewer] = field(default_factory=set)
```
trag1c: This could use a better name, but my only idea is `unrequests`.

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,283 @@
...
+    @staticmethod
+    def _format_reviewers(
+        reviewers: set[Reviewer], unknown: int = 0, *, heading: str | None = None
+    ) -> str:
+        heading_ = f"**{heading}** " * (heading is not None)
+        match len(reviewers) + bool(unknown):  # number of entries.
+            case 0:
+                return ""
+            case 1 if unknown:
+                return f"{heading_}from {unknown} unknown users"
```
trag1c: Maybe I don't see the reasoning behind doing it that way, but why not something like this?
```suggestion
        match len(reviewers):  # number of entries.
            case 0 if unknown:
                return f"{heading_}from {unknown} unknown users"
            case 0:
                return ""
```
Or perhaps even extracting that case outside the match:
```suggestion
        if not reviewers and unknown:
            return f"{heading_}from {unknown} unknown users"
        match len(reviewers):  # number of entries.
            case 0:
                return ""
```
If you went with either option, you could also consider matching against a list as a follow-up (diff assumes the second one):
```diff
-match len(reviewers) + bool(unknown):  # number of entries.
+match reviewers_ := list(reviewers):
-    case 0:
+    case []:
         return ""
-    case 1:
+    case [reviewer]:
-        reviewer = next(iter(reviewers))
         the_ = "the " * isinstance(reviewer, GitHubTeam)
         return f"{heading_}from {the_}{reviewer.format()}"
     case _:
-        asc_reviewers = sorted(reviewers, key=lambda r: r.name)
+        reviewers_.sort(key=lambda r: r.name)
         return (
             f"{heading_}from:\n"
-            + "\n".join(f"- {r.format()}" for r in asc_reviewers)
+            + "\n".join(f"- {r.format()}" for r in reviewers_)
             + (f"\n- {unknown} unknown users" if unknown else "")
         )
```

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,283 @@
...
+            case 1:
+                reviewer = next(iter(reviewers))
+                the_ = "the " * isinstance(reviewer, GitHubTeam)
```
trag1c: Do we use `GitHubTeam.format` anywhere else? If not, I'd consider throwing the `the ` in there to avoid doing logic here.

trag1c (changes requested):
Exciting!! I still haven't fully understood the core flow but it feels good overall. Have many questions =)

tests/webhooks/utils.py
00-kat: Yeah I guess it indeed can… I just did what the documentation did lol; I'll change it.

tests/webhooks/utils.py
00-kat: I personally prefer `min_value=0` or `1` instead of those aliases because “non-negative” vs “positive” is, in my opinion, an *extremely* precise and minute semantic difference.

tests/webhooks/test_review_summary.py
00-kat: I swear I typed that right, I wonder if I accidentally hit `x` (Vim for “delete letter under cursor”) while reading the code around it or something…

tests/webhooks/test_review_summary.py
00-kat: Nuke all of them and the execution time on my machine was ~7.5s on average, while with them it's 3s. These run unconditionally in the justfile's `check` rule so I felt like longer tests would be annoying.

(I initially aimed for ≤2s, but wasn't able to get it that low.)

Oh and for context:
- before this PR, the toolbox tests took 1.2s on my machine, and they're ~2s now.
- before this PR, the main app's tests took 1s on my machine, and they're 3s now.

Oh and one more edit: lowering those values much more didn't really help much from testing, it was only around 25–100ms, but running all 100 was noticeably slower.

tests/webhooks/test_review_summary.py
00-kat: > including the logs in the cases to test that as well

I wasn't sure of how that would be done. I'm not actually sure why I added those, I'll remove them unless you know of some reasonable way to assert logs.

tests/webhooks/test_review_summary.py
00-kat: I tried, it looks identical to the `__bool__` function, except in a `.filter()` 😅.

packages/toolbox/src/toolbox/misc.py
00-kat: I can't think of where it'd be useful, and it gives an allure of asynchronous execution (since each individual element would still be computed without await points). Lists are already computed, so you wouldn't have any chance of thinking it's for performance or anything; maybe `Sequence` might be more appropriate for that though?

app/components/github_integration/webhooks/review_summary.py
00-kat: `review_pools` is the dictionary in question, with the `review_pools[key] = q` being the dictionary set. I'll move it closer to the line in question though, I initially wrote it before that logging statement hogged up three whole lines.

app/components/github_integration/webhooks/review_summary.py
00-kat: Agreed, I hate all the names in this class (especially `null_set`/`opposite_null_set`), but I couldn't think of anything so just added comments instead.

`unrequests` is quite literally what it is, but I feel like it doesn't highlight the semantic difference between `unrequests` and `rerequests` that well (which might make the strings used in `format()` more confusing).

app/components/github_integration/webhooks/review_summary.py
00-kat: `case _:` right below has no `the `.

I.e.:

> from the `macos` team.

and

> from:
> - [`trag1c`](https://github.com/trag1c)
> - `macos` team

packages/toolbox/tests/utils.py
00-kat: …apparently not? Both `st.from_type(object).example()` and `st.from_type(type).flatmap(st.from_type).example()` generate stuff that looks mostly identical to my eyes at least; I wonder why their documentation mentions `st.from_type(type).flatmap(st.from_type)` multiple times and [considers it a “recipe”](https://hypothesis.readthedocs.io/en/latest/reference/strategies.html#hypothesis.strategies.SearchStrategy.flatmap) too.

tests/webhooks/test_review_summary.py
trag1c: Normally you'd do this with the built-in `caplog` fixture, but since we're using loguru you'd need to override it (based on [loguru docs](https://loguru.readthedocs.io/en/stable/resources/migration.html#replacing-caplog-fixture-from-pytest-library)):
```py
import pytest

@pytest.fixture
def caplog(caplog: pytest.LogCaptureFixture):
    handler_id = logger.add(
        caplog.handler,
        format="{message}",
        level=0,
        filter=lambda record: record["level"].no >= caplog.handler.level,
        enqueue=False,  # Set to 'True' if your test is spawning child processes.
    )
    yield caplog
    logger.remove(handler_id)
```
```py
# test_foo.py
from loguru import logger

def test_log(caplog):
    logger.warning("bar")
    assert caplog.record_tuples == [("test_foo", logging.WARNING, "bar")]
```

packages/toolbox/tests/utils.py
trag1c: FWIW their internals use both for some reason:
```py
# hypothesis/internal/conjecture/provider_conformance.py
def run_conformance_test(
    ...
    _realize_objects: SearchStrategy[Any] = (
        st.from_type(object) | st.from_type(type).flatmap(st.from_type)
    ),
) -> None:
```

app/components/github_integration/webhooks/review_summary.py
trag1c: Oh, this is "dictionary set" as in "setting a value for a dictionary key" and not "a `set` of `dict`s".

app/components/github_integration/webhooks/review_summary.py
trag1c: I feel like it'd be reasonable to use `the ` there 🤔

app/components/github_integration/webhooks/review_summary.py
00-kat: LOL yeah. Would calling it “dictionary update” help?

app/components/github_integration/webhooks/review_summary.py
00-kat: My only reason for not including it is because the code blocks align really nicely without them.

app/components/github_integration/webhooks/review_summary.py
trag1c: Yes, definitely haha

app/components/github_integration/webhooks/review_summary.py
trag1c: Ah, yeah, that is true.

tests/webhooks/utils.py
00-kat: I don't think I like this all that much honestly; almost all other ones are functions which get called, often with no arguments, so having a bunch need the `()` while others don't for no particular reason but simplified implementation is kinda arbitrary and annoying to remember. I can change it if you strongly believe this is nicer but I'm marking this as resolved for now ^^.

packages/toolbox/src/toolbox/misc.py
00-kat: I generally only add backticks if it's not obvious that it's referring to an identifier or code; in that example it isn't immediately clear that the lack of grammar around `mock` isn't just a typo. I'm fine with just being more uniform with these though.

app/components/github_integration/webhooks/review_summary.py
00-kat: I think I'll just delete that line actually, since everything else assumes the GIL exists anyway so it just feels like a confusing extra detail that somehow matters more here.

app/components/github_integration/webhooks/review_summary.py
00-kat: Do you have any particular opinion on the `the` in that case? I don't care too much and amn't sure if I mark this as resolved or not.

packages/toolbox/src/toolbox/misc.py
00-kat: One thing I just realized: not having the backquotes consistently makes it harder to know what *isn't* an identifier 😅 I guess I changed my mind on that practice of mine entirely. I realized this because `event` in the next comment wasn't trying to point to the variable specifically, but you had assumed it was referring to the variable (as your suggestion wrapped it in backquotes), and it's far clearer what I intended to be variable references with the explicit backquotes.

tests/webhooks/test_review_summary.py
00-kat: I guess I could have a few hardcoded unit tests instead of trying to have a property test, but I'm not really sure what values to test… Just doing a few random small values with no rhythm or rhyme to it feels wrong to me for some inexplicable reason.

packages/toolbox/src/toolbox/misc.py
00-kat: Yeah looks like a `Sequence` is most appropriate (per a reading of the docs); I'll switch to that and add a comment about generalizing this to `Iterable`.

app/components/github_integration/webhooks/review_summary.py
00-kat: > Maybe I don't see the reasoning behind doing it that way

The reasoning is that I directly translated what I was thinking as I tried to clean up my previous if-else mess while handling all edge cases into code :P.

> you could also consider matching against a list

This was my original intent with the `match` actually, but for some reason I didn't want to convert it to a list once I found out I can't Just Do That™ with a set. Doesn't apply anymore though, since the `sorted` was a last-minute addition that makes it a list anyway when I realized I completely forgot about it.

00-kat: Oh and: both your ways don't handle the case when there's one reviewer and non-zero unknowns, I think. (You'd go down the `len(reviewer) == 1` case, formatting the reviewer but completely discarding the unknowns, while mine would go down the `case _:`.) After trying to do it with `len(reviewers)` I gave up and expressed what I was thinking, which is that unknowns is a separate line so should be counted alongside the `len(reviewers)`. I'll switch it to pattern matching on a list though, and somehow incorporate the unknown handling into it (probably with an auxiliary `if` guard).

Edit: I didn't trust my thinking and wanted to make sure, so I applied your diffs and ran the tests, and sure enough, all the unknown handling tests exploded loudly 😅. Too much text for me to include here, feel free to run them yourself if you need convincing of what I said too.

00-kat: > probably with an auxiliary `if` guard

This was my diff; I don't love it because there's two `if`s for unknown handling now, but this did give me an easy way to add a comment explaining it (as opposed to the previous version, where I'd have to explain in a large paragraph why thinking of it abstractly as “lines” instead of “reviewers + maybe unknown” is useful for cleanly branching on its count). You might be able to think up something “cleaner”.

```diff
         heading_ = f"**{heading}** " * (heading is not None)
-        match len(reviewers) + bool(unknown):  # number of entries.
+        match sorted(reviewers, key=lambda r: r.name):
-            case 0:
+            case [] if unknown:
+                # Only unknowns.
+                return f"{heading_}from {unknown} unknown users"
+            case []:
                 return ""
-            case 1 if unknown:
-                return f"{heading_}from {unknown} unknown users"
-            case 1:
-                reviewer = next(iter(reviewers))
+            # Unknowns go on a separate line, so if there is both a reviewer and
+            # unknowns, go down the last case instead so that they're formatted on
+            # multiple lines.
+            case [reviewer] if not unknown:
                 the_ = "the " * isinstance(reviewer, GitHubTeam)
                 return f"{heading_}from {the_}{reviewer.format()}"
-            case _:
-                asc_reviewers = sorted(reviewers, key=lambda r: r.name)
+            case rs:
                 return (
                     f"{heading_}from:\n"
-                    + "\n".join(f"- {r.format()}" for r in asc_reviewers)
+                    + "\n".join(f"- {r.format()}" for r in rs)
                     + (f"\n- {unknown} unknown users" if unknown else "")
                 )
 
```

tests/webhooks/test_review_summary.py
00-kat: Honestly I think I'll just drop the comments, I don't want to do all the special-casing needed to test it properly (pulling it out to a separate test would make for a bunch more boilerplate, especially around explanation comments about why the tests are shared, and it being in a different order than I intended it to be in; making it optional or something would result in special-casing in all the test functions; incorporating the special casing in the helpers would be annoying because the parallel one can't have this tested so that needs to be dropped somehow; I don't want to figure out what magic incantation I need to configure the fixture's log level; and finally I don't want to figure out how ~~Java-style pure magic dependency injection framework hell~~fixtures work in the presence of helper functions).

00-kat:
> core flow

In case it helps, the review request events generally work like this now:

1. The webhook hook (silly name) is called on every event, and it delegates to `handle_review_request` until it eventually returns a non-falsy result (upon which we know we have a summary available and ready to send).
2. `handle_review_request` is finicky reentrant queue stuff to synchronize and segregate raw review request events into PRs review operation clusters, converting it to an asynchronous iterator and dealing with the timeout. It then simply passes that aiter to `ReviewRequestSummary.collect`. If it weren't for the timeout, I'd just call this “annoying synchronization glue”.
3. `ReviewRequestSummary.collect` is the actual logic, this is probably the most confusing part. It's helpful to think of this mini state machine as a singular “transaction” (semantic request/removal group, à la checking or unchecking reviewers in GitHub's UI in one go), where the only important states are the world before these requests and the world after the timeout, since that's what the function does: collect multiple, *ordered* review requests into a singular review update transaction that reflects the difference between the initial and final states; that would help clarify some of the cases in `_categorize`, especially the `r in null_set` one[^1]. Also, keep in mind that a reviewer would alternate between `default_set`/`null_set` and `opposite_set`/`opposite_null_set` outside of GitHub being trash, because they're alternating requests and removals. Finally, stepping through the states manually (e.g. by executing the tests by hand) is helpful.
4. Back at the hook, once we have a `ReviewRequestSummary`, we simply `.format()` it and pass it to that embed helper that every other hook uses.
5. Said `.format()` function is extremely intricate logic and special-casing solely to get nicer formatted output, but otherwise a straightforward pure function from `ReviewRequestSummary` to title (`str`) + body (`str`).

[^1]: This actually initially just showed the last two states, since these can cycle around indefinitely and we had to cut it off *somewhere* lest the formatting be impractical. I only arrived at the “transaction” analogy after a couple days of looking at that function and repeatedly forgetting to refactor it (it originally duplicated the code for the requests and removals, instead of using a `_categorize` function, because I initially wrote the entire logic in an hour or two entirely inline in `register_hooks`, and picking parameter names is hard 😄).

app/components/github_integration/webhooks/review_summary.py
trag1c: I'd say throw the `the ` into the repr, the nice alignment will be lost but (1) the list will never be huge enough for it to matter at all (2) it'll simplify stuff quite a bit.

app/components/github_integration/webhooks/review_summary.py
trag1c: Honestly I think I'd find fun names like `oops_requests` (for `no_requests`) and `oops_removals` (for `rerequests`) more understandable 😄

00-kat: I still feel like this holds for those:

> `unrequests` is quite literally what it is, but I feel like it doesn't highlight the semantic difference between `unrequests` and `rerequests` that well (which might make the strings used in `format()` more confusing).

(Slightly less fun names would be `accidental_...`, lol.)

tests/webhooks/test_review_summary.py
trag1c: Mmm yeah scratch that then.

app/components/github_integration/webhooks/review_summary.py
00-kat: …oh wait, `accidental_requests` and `rerequests` works. The main problem is it calls them accidental “requests” when they correspond to `removals`, but I feel like it's still a bit better than `no_requests`, and also matches what the format function calls them.

trag1c: I feel like the most explicit names (that would let you drop the explainer comments) would (unfortunately) be `requested_then_removed` and `removed_then_requested` (with the others becoming `requested` and `removed` to be grammatically consistent).

Also asked an LLM to spit something out and it suggested `requests`, `removals`, `retracts`, and `restores`, not sure what to think about these.

00-kat: The main reason I dislike the explicit version is the tense and plurality. The former can be `removed_requests` but I'm stuck on the latter (`requested_removals` is obviously wrong lmao), and even `removed_requests` is confusing because `removals` also tracks requests that were removed.

I think retracts and restores are even worse because neither makes it obvious which it's the “undo” of lol.

app/components/github_integration/webhooks/review_summary.py
```diff
@@ -0,0 +1,284 @@
...
+    event: ReviewRequestsModified,
+    *,
+    timeout: int | None = REVIEW_REQUEST_GROUP_TIMEOUT,  # noqa: ASYNC109
+    force_shutdown: bool = False,
+) -> ReviewRequestSummary | None:
+    """
+    `force_shutdown` can be set to forcibly shut down the pool associated with this
+    event's pull request and return a review summary immediately, bypassing the timeout;
+    this is primarily for testing.
+    """
```
trag1c: It would be great to not have to introduce a parameter that's only useful for testing.

<details><summary>I asked an LLM to suggest a different approach and it spat out something like this (which looks reasonable at first glance, but I think you'll ultimately be the best judge as you're the most familiar with the code):</summary>

```diff
diff --git a/app/components/github_integration/webhooks/review_summary.py b/app/components/github_integration/webhooks/review_summary.py
index 536f25a..b2e9865 100644
--- a/app/components/github_integration/webhooks/review_summary.py
+++ b/app/components/github_integration/webhooks/review_summary.py
@@ -19,7 +19,23 @@ type Reviewer = GitHubUser | GitHubTeam
 type ReviewRequestsModified = (
     events.PullRequestReviewRequested | events.PullRequestReviewRequestRemoved
 )
-type ReviewPools = dict[ReviewPoolKey, asyncio.Queue[ReviewRequestsModified]]
+class ReviewPool:
+    """Encapsulates a queue of review request events and their collection."""
+
+    def __init__(self) -> None:
+        self.queue = asyncio.Queue[ReviewRequestsModified]()
+
+    async def put(self, event: ReviewRequestsModified) -> None:
+        await self.queue.put(event)
+
+    def shutdown(self) -> None:
+        self.queue.shutdown()
+
+    async def collect(self) -> ReviewRequestSummary:
+        return await ReviewRequestSummary.collect(drain_queue(self.queue))
+
+
+type ReviewPools = dict[ReviewPoolKey, ReviewPool]
 
 
 class ReviewPoolKey(NamedTuple):
@@ -37,56 +53,35 @@ async def handle_review_request(
     event: ReviewRequestsModified,
     *,
     timeout: int | None = REVIEW_REQUEST_GROUP_TIMEOUT,  # noqa: ASYNC109
-    force_shutdown: bool = False,
 ) -> ReviewRequestSummary | None:
-    """
-    `force_shutdown` can be set to forcibly shut down the pool associated with this
-    event's pull request and return a review summary immediately, bypassing the timeout;
-    this is primarily for testing.
-    """
     key = ReviewPoolKey(pr_number=event.pull_request.number, actor_id=event.sender.id)
 
-    # If we already have a queue for this PR+user combination, simply collect to it.
-    if q := review_pools.get(key):
+    # If we already have a pool for this PR+user combination, simply add to it.
+    if pool := review_pools.get(key):
         logger.info("found existing review pool for key {key}", key=key)
-        if force_shutdown:
-            logger.warning(
-                "forcibly shutting down review pool {key} ({n} remain)",
-                key=key,
-                n=len(review_pools) - 1,
-            )
-            del review_pools[key]
-            q.shutdown()
-        else:
-            await q.put(event)
+        await pool.put(event)
         return None
 
     logger.info(
         "creating new review pool #{n} for key {key}", n=len(review_pools) + 1, key=key
     )
-    q = asyncio.Queue[ReviewRequestsModified]()
-    review_pools[key] = q
-    # This won't throw QueueShutDown as it cannot be shut down yet: the timeout
-    # hasn't been set.
-    await q.put(event)
-    # The coroutine calling this function will see this queue through to the end;
-    # future review_request() calls for this queue will terminate immediately after
-    # updating the queue.
+    pool = ReviewPool()
+    review_pools[key] = pool
+    await pool.put(event)
 
     async def timer() -> None:
         if timeout is None:
             return
         await asyncio.sleep(timeout)
         del review_pools[key]
-        # Signal the timeout to the queue.
-        q.shutdown()
+        pool.shutdown()
         logger.info(
             "shut down review pool {key} ({n} remain)", key=key, n=len(review_pools)
         )
 
     async with asyncio.TaskGroup() as group:  # transparently handle cancelation.
         group.create_task(timer())
-        summary = group.create_task(ReviewRequestSummary.collect(drain_queue(q)))
+        summary = group.create_task(pool.collect())
 
     return summary.result()
 
diff --git a/tests/webhooks/test_review_summary.py b/tests/webhooks/test_review_summary.py
index f064a95..63c71ff 100644
--- a/tests/webhooks/test_review_summary.py
+++ b/tests/webhooks/test_review_summary.py
@@ -25,6 +25,8 @@ from tests.webhooks.utils import (
 from app.components.github_integration.models import GitHubTeam, GitHubUser
 from app.components.github_integration.webhooks.review_summary import (
     Reviewer,
+    ReviewPool,
+    ReviewPoolKey,
     ReviewPools,
     ReviewRequestsModified,
     ReviewRequestSummary,
@@ -270,14 +272,14 @@ async def test_collect_review_request_summary(
     assert await ReviewRequestSummary.collect(seq_to_aiter(requests)) == result
 
 
-# Since handle_review_request is just synchronization glue around
+# Since ReviewPool is just synchronization glue around
 # ReviewRequestSummary.collect, the same tests should produce the same results.
 @pytest.mark.parametrize(("requests", "result"), REVIEW_REQUEST_COLLECTION_TESTS)
-async def test_handle_review_request_collection(
+async def test_review_pool_collection(
     requests: list[Mock], result: ReviewRequestSummary
 ) -> None:
     pools: ReviewPools = {}
-    await run_review_request_collection_test(pools, 12, requests, result)
+    await run_pool_collection_test(pools, 12, requests, result)
 
 
 # Same as the test above, except in parallel, to catch cases where having multiple PRs
@@ -286,18 +288,18 @@ async def test_handle_review_request_collection(
 # run instead of only this one, despite this one being a superset of the other one.
 @settings(max_examples=10)
 @given(st.permutations(REVIEW_REQUEST_COLLECTION_TESTS))
-async def test_handle_review_request_parallel_collection(
+async def test_review_pool_parallel_collection(
     tests: list[tuple[list[Mock], ReviewRequestSummary]],
 ) -> None:
     pools: ReviewPools = {}
     async with asyncio.TaskGroup() as group:
         for pr, (requests, result) in enumerate(tests, 1):
             group.create_task(
-                run_review_request_collection_test(pools, pr, requests, result)
+                run_pool_collection_test(pools, pr, requests, result)
             )
 
 
-async def run_review_request_collection_test(
+async def run_pool_collection_test(
     pools: ReviewPools,
     pr_number: int,
     requests: list[Mock],
@@ -305,27 +307,49 @@ async def run_review_request_collection_test(
 ) -> None:
     first, *rest = (add_pr_metadata(m, pr_number) for m in requests)
 
+    key = ReviewPoolKey(pr_number=pr_number, actor_id=42)
+    pool = ReviewPool()
+    pools[key] = pool
+
     async def enqueue() -> None:
+        await pool.put(first)
         for r in rest:
             rv = await handle_review_request(pools, r, timeout=None)
             assert rv is None
             await asyncio.sleep(0)  # context switch.
-        # Explicitly shut down the queue.
-        rv = await handle_review_request(
-            pools,
-            add_pr_metadata(Mock(()), pr_number),
-            timeout=None,
-            force_shutdown=True,
-        )
-        assert rv is None
+        pool.shutdown()
 
     async with asyncio.TaskGroup() as group:
-        summary = group.create_task(handle_review_request(pools, first, timeout=None))
+        summary = group.create_task(pool.collect())
         group.create_task(enqueue())
 
     assert summary.result() == result
 
 
+async def test_handle_review_request_creates_pool() -> None:
+    pools: ReviewPools = {}
+    event = add_pr_metadata(user_review_request(PRReviewReq, name="foo"), 1)
+
+    summary = await handle_review_request(pools, event, timeout=0)
+    assert summary == ReviewRequestSummary(requests={dummy_user(name="foo")})
+    assert len(pools) == 0
+
+
+async def test_handle_review_request_uses_existing_pool() -> None:
+    pools: ReviewPools = {}
+    key = ReviewPoolKey(pr_number=1, actor_id=42)
+    pool = ReviewPool()
+    pools[key] = pool
+
+    event = add_pr_metadata(user_review_request(PRReviewReq, name="foo"), 1)
+    rv = await handle_review_request(pools, event, timeout=None)
+    assert rv is None
+
+    pool.shutdown()
+    summary = await pool.collect()
+    assert summary == ReviewRequestSummary(requests={dummy_user(name="foo")})
+
+
 @given(st.none() | st.text())
 def test_format_reviewers_empty(heading: str | None) -> None:
     assert ReviewRequestSummary._format_reviewers(set(), 0, heading=heading) == ""

```
</details>

00-kat: The problems with the LLM's solution:

1. Shutting down doesn't test `handle_review_request`, it tests `ReviewPool.shutdown()`. What if `handle_review_request` accidentally does `q.shutdown()` instead of `review_pools.shutdown()`?
2. `ReviewPool` delegates to `Queue`, so I could even now get the pool myself and call `.shutdown()` on it. It's really just a type-restricted `Queue` with a helper method, there isn't much of a point.
3. Shutting down currently does:
   > ```py
   > del review_pools[key]
   > q.shutdown()
   > ```
   While the LLM's suggestion does:
   > ```py
   > q.shutdown()
   > ```

(2) and (3) are easily solvable by making a new class to store all review pools and including a shutdown method on it that takes the key instead of a type alias to a dictionary, but it doesn't solve (1) and I don't like creating boilerplate classes. (**Edit**: one other thing is that the shutdown is now conceptually on the pools, and yeah sure the OOP way to think of it is that you have a chat with the pool registrar and tell it to decide to shutdown its pool then forget about it, but I don't think shutting down should be considered a method of a pool registrar instead of a pool unless being more OOP matters—in which case we should be dealing with the pools solely via its registrar.)

I don't like parameters solely for testing either, but I can't think of any other way that solves (1). Actually I've done these kwonly `force_*` parameters before too for testing: https://github.com/ghostty-org/discord-bot/blob/aa76254c50e6b374dd2a257935a5f7b241c035af/packages/toolbox/src/toolbox/message_moving/conversion.py#L38-L44 That one doesn't look nearly as egregious because it's not literally used in the tests, but it's to work around me not having Nitro 😅.

00-kat: Also, re the two new tests: I was actually very hopeful with `test_handle_review_request_creates_pool`, because I didn't know how to test that the pools are correct without going insane lol, but it seems to test… nothing? Lmao. And `test_handle_review_request_uses_existing_pool` is closer to what I wanted to test, but once you drain the queue in the `collect`, you can't run `handle_review_request` again (it'll error with `QueueShutDown`, I believe), so this simply tests that *a* pool is used, not that an *existing* pool is used by subsequent calls, which in my opinion is the easier thing to get wrong.

The reason I say this is: do you know how I could test that the pools are used correctly while still being able to use `pytest.mark.parametrize` and with sane code? This is of course tangential to this review thread but I never thought to ask if you had any ideas.

00-kat: > What if `handle_review_request` accidentally does `q.shutdown()` instead of `review_pools.shutdown()`?

Okay hm, this is possible even now, since `timeout` and `force_shutdown` simply duplicate the same code. I could pull that out into a helper closure I guess, but it's still possible to forget to use it, so my argument there isn't really valid.

I guess I could just use `.shutdown()` on the queue in the test. I'm not actually sure why I assert that the function returns `None` afterward, since when `timeout` shuts it down it would actually return the pools, and anyway this doesn't test the timeout.

app/components/github_integration/webhooks/review_summary.py
00-kat: Honestly I think this has been bikeshed enough and I'm doubtful we'll come up with anything better, so I'm going to go with the following, and if in the future someone has a mysterious revelation we can change it then :P.

```py
requests
removals
accidental_requests  # When a review was requested from somebody, then later removed.
rerequests  # When a review request for somebody was removed, then later re-requested.
```

trag1c: Sounds good!

app/components/github_integration/webhooks/review_summary.py
00-kat: I do agree with (1), and given that the list won't be large enough for it to matter I guess the `the` makes for more fluent grammar.

I don't agree with (2) though; here's the diff:
```diff
diff --git a/app/components/github_integration/models.py b/app/components/github_integration/models.py
index 5d70d2003c..4a01157496 100644
--- a/app/components/github_integration/models.py
+++ b/app/components/github_integration/models.py
@@ -53,7 +53,7 @@
     name: str
 
     def format(self) -> str:
-        return f"`{self.name}` team"
+        return f"the `{self.name}` team"
 
 
 class Reactions(BaseModel, frozen=True):
diff --git a/app/components/github_integration/webhooks/review_summary.py b/app/components/github_integration/webhooks/review_summary.py
index 043fe4cbe0..56c3865c7c 100644
--- a/app/components/github_integration/webhooks/review_summary.py
+++ b/app/components/github_integration/webhooks/review_summary.py
@@ -239,8 +239,7 @@
             # unknowns, go down the last case instead so that they're formatted on
             # multiple lines.
             case [reviewer] if not unknown:
-                the_ = "the " * isinstance(reviewer, GitHubTeam)
-                return f"{heading_}from {the_}{reviewer.format()}"
+                return f"{heading_}from {reviewer.format()}"
             case rs:
                 return (
                     f"{heading_}from:\n"
```
And the diff for the tests:
```diff
diff --git a/tests/webhooks/test_review_summary.py b/tests/webhooks/test_review_summary.py
index e5936b0923..54976f715f 100644
--- a/tests/webhooks/test_review_summary.py
+++ b/tests/webhooks/test_review_summary.py
@@ -463,1 +463,1 @@
-            "from:\n- [`a`](<>)\n- `b` team\n- 6 unknown users",
+            "from:\n- [`a`](<>)\n- the `b` team\n- 6 unknown users",
@@ -504,1 +504,1 @@
-            "from:\n- [`a`](<>)\n- `b` team\n- 3 unknown users",
+            "from:\n- [`a`](<>)\n- the `b` team\n- 3 unknown users",
@@ -521,1 +521,1 @@
-            "**Accidentally requested review** from:\n- `a` team\n- [`b`](<>)",
+            "**Accidentally requested review** from:\n- the `a` team\n- [`b`](<>)",
@@ -538,1 +538,1 @@
-            "**Removed, then requested review** from:\n- [`a`](<>)\n- `b` team",
+            "**Removed, then requested review** from:\n- [`a`](<>)\n- the `b` team",
@@ -661,1 +661,1 @@
-            "**Removed review request** from:\n- [`a`](<>)\n- `b` team"
+            "**Removed review request** from:\n- [`a`](<>)\n- the `b` team"
@@ -672,1 +672,1 @@
-            "**Removed review request** from:\n- `a` team\n- [`b`](<>)\n\n"
+            "**Removed review request** from:\n- the `a` team\n- [`b`](<>)\n\n"
@@ -687,1 +687,1 @@
-            "**Requested review** from:\n- [`a`](<>)\n- `b` team\n- `c` team"
+            "**Requested review** from:\n- [`a`](<>)\n- the `b` team\n- the `c` team"
@@ -690,2 +690,2 @@
-            "**Accidentally requested review** from:\n- [`d`](<>)\n- `e` team\n\n"
-            "**Removed, then requested review** from:\n- `q` team\n- `r` team",
+            "**Accidentally requested review** from:\n- [`d`](<>)\n- the `e` team\n\n"
+            "**Removed, then requested review** from:\n- the `q` team\n- the `r` team",
```
The syntax highlighting doesn't make it obvious, but each line had a `the` or two inserted.

trag1c (approved):
Let's try this out, great work!!

PR #546, opened 2026-05-07
Disallow moving messages that fail the destination's message filters

### LINKED ISSUE #542: Disallow moving a message if it wouldn't pass the new location's message filters
If you send `a`, then move it to `#showcase`, it gets deleted immediately with three “attempted to DM a bot” warnings. Instead, selecting `#showcase` should show something akin to what happens with system messages:

Screenshot of kat's Ghostty Bot instance responding with “System messages cannot be moved”, and providing a “❌️ Delete instead” button.

<br />

I believe there's a `DeleteInstead` view class or something used by system messages that can be repurposed for this.

Note that messages which pass the message filter—i.e., moving images to `#showcase` or links to `#media`—should be allowed; it's only those that would fail the message filter that should get the delete instead treatment.

Maybe adding a “Try again”/“Select another channel” button alongside “Delete instead” might be useful; I'm not sure if it's likely that someone would accidentally press `#showcase`/`#media` though.

00-kat:
Closes #542.

### DIFF
--- a/app/components/message_filter.py
+++ b/app/components/message_filter.py
@@ -4,7 +4,7 @@ import discord as dc
 from discord.ext import commands
 
 from app.config import config
-from toolbox.discord import format_or_file, try_dm
+from toolbox.discord import GuildTextChannel, format_or_file, try_dm
 from toolbox.messages import REGULAR_MESSAGE_TYPES
 from toolbox.misc import URL_REGEX
 
@@ -14,8 +14,8 @@ if TYPE_CHECKING:
     from app.bot import GhosttyBot
 
 _MESSAGE_DELETION_TEMPLATE = (
-    "Hey! Your message in {} was deleted because it did not contain {}. "
-    "Make sure to include {}, and respond in threads.\n"
+    "Hey! Your message in {channel} was deleted because it did not contain "
+    "{requirement}. Make sure to include {suggestion}, and respond in threads.\n"
 )
 _MESSAGE_CONTENT_NOTICE = "Here's the message you tried to send:"
 _COPY_TEXT_HINT = (
@@ -27,7 +27,8 @@ _COPY_TEXT_HINT = (
 class MessageFilterTuple(NamedTuple):
     channel_id: int
     filter: Callable[[dc.Message], object]
-    template_fillers: tuple[str, str]
+    requirement: str
+    suggestion: str
 
 
 @final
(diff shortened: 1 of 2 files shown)


app/components/message_filter.py
```diff
@@ -57,10 +60,18 @@ def check(self, message: dc.Message) -> MessageFilterTuple | None:
...
         """
+        assert isinstance(message.channel, GuildTextChannel)
+        return self.check_in(message.channel, message)
+
+    def check_in(
+        self, channel: GuildTextChannel, message: dc.Message
+    ) -> MessageFilterTuple | None:
+        """
+        Same as check(), but tries the message filters of the specified channel rather
+        than the channel the message is in.
```
CoderJoshDK: The comment doesn't actually make sense. Since `check` just calles `check_in`, it shouldn't really reffernce `check`. That should be the other way around.

CoderJoshDK (commented):
Super tiny nit. The logic itself seems good

trag1c (approved):
Looks clean! Will merge after Joshie's comment is resolved.

app/components/message_filter.py
00-kat: I felt that `check_in` was easier to describe in terms of `check`, despite `check` being the special case of `check_in`; this docstring was moreso intended for people looking at it with an LSP or whatever.

CoderJoshDK: The copy of each docstring can be very similar to eachother if you want. You don't have to pure collapse all of `check` to be "Calls `check_in` for the channel the message is in." You can write out the whole thing twice. And I would rather that. IMO calling another function is just an implementation detail. The LSP docstring shouldn't care about that.

00-kat: > IMO calling another function is just an implementation detail. The LSP docstring shouldn't care about that.

Exactly, which is why I didn't care about that: `check` calls `check_in`, but the docstring of `check_in` describes itself in terms of `check`; i.e.: the exact opposite of the source code. (Contrary to your review comment which was concerned with who calls who in the source code.)

> You can write out the whole thing twice.

Fair enough, I'll see if I can word it well without sounding like a copy of `check`'s docstring with an extra clause shoved on.

CoderJoshDK: You are the resident word smith, so I will default to your final decision.

> Exactly, which is why I didn't care about that: `check` calls `check_in`, but the docstring of `check_in` describes itself in terms of `check`

My main comment is really trying to point out that the docstring, as is, feels inaccurate. That is why I am opposed to `check_in` being in terms of `check`.

00-kat: Fair enough. See if you like my new docstring (which I too felt is more accurate as to the purpose of `check_in` after writing it).

CoderJoshDK (approved):

PR #547, opened 2026-05-08
Indicate that attachments are being uploaded when moving a message

### LINKED ISSUE #545: Moving messages shows no indication that the request was received
The thinking ellipses is removed as soon as the bot receives the channel selection (and defers the response). However, if there are attachments (either directly or in the forward), Ghostty Bot might take a long time before it actually moves it, but there is no indication that Ghostty Bot didn't just break. Past me apparently decided to add an indication when editing attachments into a message: https://github.com/ghostty-org/discord-bot/blob/04302f51a3367eb9d914c35c7fc33f465690553b/app/components/move_message.py#L113 This message should show up when moving a message (or converting it to a help post) with either an attachment, or a forward containing an attachment, as well.

00-kat:
Closes #545.

### DIFF
--- a/app/components/move_message.py
+++ b/app/components/move_message.py
@@ -183,9 +183,9 @@ class SelectChannel(SafeView):
         max_values=1,
     )
     async def select_channel(
-        self, interaction: dc.Interaction, sel: dc.ui.ChannelSelect[Self]
+        self, interaction: dc.Interaction, select: dc.ui.ChannelSelect[Self]
     ) -> None:
-        channel = await self.bot.fetch_channel(sel.values[0].id)
+        channel = await self.bot.fetch_channel(select.values[0].id)
         assert isinstance(channel, GuildTextChannel)
         if channel.id == self.message.channel.id:
             await interaction.response.edit_message(
@@ -211,7 +211,15 @@ class SelectChannel(SafeView):
             )
             return
 
-        await interaction.response.defer()
+        select.disabled = True
+        if self.message.attachments or any(
+            m.attachments for m in self.message.message_snapshots
+        ):
+            await interaction.response.edit_message(content=_UPLOADING, view=self)
+        else:
+            await interaction.response.defer()
+            await interaction.edit_original_response(view=self)
+
         webhook_channel, thread = (
             (channel.parent, channel)
             if isinstance(channel, dc.Thread)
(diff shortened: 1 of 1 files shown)


app/components/move_message.py
```diff
@@ -166,7 +166,13 @@ async def select_channel(
             )
             return
 
-        await interaction.response.defer()
+        if self.message.attachments or any(
+            m.attachments for m in self.message.message_snapshots
+        ):
+            await interaction.response.edit_message(content=_UPLOADING)
```
CoderJoshDK: Should we remove the view here? It would simplify a few things (not code wise). However, on the off chance that the move fails, the interface wouldn't let you retry or chose something different. So,,, thoughts?

00-kat: If the move fails, the class is probably in some inconsistent state anyway (so retrying would just break more things). Most other cases disable their buttons but keep the view there; I'll see if it's possible to disable a dropdown and if not I'll just delete the view.

00-kat: Turns out, this bug exists on main already 😅.

Screenshot of high quality testing in production.

Good job on noticing it!

app/components/move_message.py
```diff
@@ -166,7 +166,15 @@ async def select_channel(
             )
             return
 
-        await interaction.response.defer()
+        sel.disabled = True
```
CoderJoshDK: I know this isn't the scope of this PR - you can make this change in another PR, hence the approval - but do we really want to call this `sel`? It was already. But I really thought you just typo'd `self` and was really confused for a second about why the CI didn't fail here. 
We can be verbose on names, that is ok. Something like `selector` or something.

CoderJoshDK (approved):

app/components/move_message.py
00-kat: It's `dc.ui.select` and `ChannelSelect` so I decided to go with `select`. No opinion on changing it to `selector` if you like that more.

app/components/move_message.py
```diff
@@ -166,7 +166,15 @@ async def select_channel(
...
+        if self.message.attachments or any(
+            m.attachments for m in self.message.message_snapshots
+        ):
```
trag1c, line 217: How do you feel about getting a property/method for this entire condition?

00-kat: Property of what exactly? I'd prefer to not have more `ExtensibleMessage`s if I could avoid it.

I thought of making it a function in toolbox but it felt too small to justify thinking up a good name for it lol.

app/components/move_message.py
CoderJoshDK: Anything is better than `sel`. I only picked `selector` because it has no risk of name shadowing. But no strong opinion there.

app/components/move_message.py
trag1c: Ah, I missed the fact that `SelectChannel` and `HelpPostTitle` are _quite separate_; I was thinking it would feel good to unify L170–172 and L242–244 :D

CoderJoshDK (approved):

trag1c (approved):

PR #549, opened 2026-05-15
Fix “No, Keep content”

00-kat:
Oops.

### DIFF
--- a/app/components/move_message.py
+++ b/app/components/move_message.py
@@ -635,6 +635,8 @@ class AttachmentChoice(SafeView):
     async def keep(
         self, interaction: dc.Interaction, _button: dc.ui.Button[Self]
     ) -> None:
+        # Set the content to MISSING so that discord.py ignores the content argument
+        # when editing.
         await self._edit(interaction, dc.utils.MISSING)
 
     @dc.ui.button(label="Yes, discard content", emoji="🖼️")  # test: allow-vs16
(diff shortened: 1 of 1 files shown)


CoderJoshDK (approved):

app/components/move_message.py
```diff
@@ -647,11 +649,12 @@ async def apply_edit_from_thread(
         new_content: str,
     ) -> None:
         channel = moved_message.channel
-        converted_content = convert_nitro_emojis(
-            self.bot, config().ghostty_guild, new_content
-        )
-        if len(converted_content) <= 2000:
-            new_content = converted_content
+        if new_content is not dc.utils.MISSING:
```
onerandomusername, line 702: why doesn't `if new_content:` work here? 

pros:
- lack of using utils.MISSING
- works even if the library switches to using None
- ignores empty strings

00-kat: I did that initially, but switched to `is not dc.utils.MISSING` so that I don't need to include a comment like “to filter out dc.utils.MISSING”.

> - lack of using utils.MISSING
> - works even if the library switches to using None

We use it everywhere already, including for this specific function. And we even explicitly pass `dc.utils.MISSING` to this function, so it's not like `if new_content:` gains us anything w.r.t. this: https://github.com/ghostty-org/discord-bot/blob/4c3384a39f0ecd34c05ed4c69b07508c3d73c30b/app/components/move_message.py#L588-L590

> - ignores empty strings

Not really. Here's the code for reference: https://github.com/ghostty-org/discord-bot/blob/4c3384a39f0ecd34c05ed4c69b07508c3d73c30b/app/components/move_message.py#L652-L657

On an empty string, execution will look a bit like:
```py
if "" is not dc.utils.MISSING:
    # Path taken.
    converted_content = convert_nitro_emojis(self.bot, config().ghostty_guild, "") 
    # converted_content == ""
    if len("") <= 2000: 
        # 0 is ≤2000
        new_content = "" 
# At this point, new_content is "" again, which is what it started off as.
```

In both cases, `""` makes it to the edit call, which will throw an error or whatever (which it should, since passing an empty string indicates a bug).

CoderJoshDK (approved):

trag1c (approved):

PR #550, opened 2026-05-15
Use monotonic clock for edit thread timeout

00-kat:
Oops.

### DIFF
--- a/app/components/move_message.py
+++ b/app/components/move_message.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime as dt
+import time
 from contextlib import suppress
 from dataclasses import dataclass
 from functools import partial
@@ -121,7 +122,7 @@ _UPLOADING = "⌛ Uploading attachments (this may take some time)…"
 class ThreadState:
     moved_message: MovedMessage
     split_subtext: SplitSubtext
-    last_update: dt.datetime
+    last_update: float  # monotonic time in seconds.
 
 
 @final
@@ -455,7 +456,7 @@ class ChooseMessageAction(SafeView):
                 view=CancelEditing(self._cog, thread),
             )
         self._cog.edit_threads[thread.id] = ThreadState(
-            self._message, self._split_subtext, dt.datetime.now(tz=dt.UTC)
+            self._message, self._split_subtext, time.monotonic()
         )
         self._cog.edit_thread_creators[interaction.user.id] = thread.id
         await self._cog.remove_edit_thread_after_timeout(thread, interaction.user)
@@ -577,7 +578,7 @@ class ContinueEditing(SafeView):
     async def continue_editing(
         self, interaction: dc.Interaction, _button: dc.ui.Button[Self]
     ) -> None:
-        self._cog.edit_threads[self._thread.id].last_update = dt.datetime.now(tz=dt.UTC)
+        self._cog.edit_threads[self._thread.id].last_update = time.monotonic()
         assert interaction.message is not None
         await interaction.message.delete()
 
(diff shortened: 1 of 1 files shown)


CoderJoshDK (approved):
Good call

app/components/move_message.py
```diff
@@ -751,34 +752,32 @@ async def remove_edit_thread(
...
-        # don't need to pass the remaining time around.
-        elapsed = dt.datetime.now(tz=dt.UTC) - self.edit_threads[thread.id].last_update
-        time_to_warning = dt.timedelta(minutes=10) - elapsed
-        # Keep sleeping until we have a zero or negative time to the warning. This has
-        # to be a while loop because other coroutines may have touched the last update
-        # time of this edit thread while we were sleeping.
-        while time_to_warning > dt.timedelta():
-            await asyncio.sleep(time_to_warning.total_seconds())
+        ten_minutes = dt.timedelta(minutes=10).total_seconds()
+        fifteen_minutes = dt.timedelta(minutes=15).total_seconds()
```
trag1c: Can we make these traditional int mul consts instead?
```py
TEN_MINUTES = 10 * 60
FIFTEEN_MINUTES = 15 * 60
```
Or, at the very least, change the current form to consts.

trag1c (commented):
Good catch, one nit though :)

app/components/move_message.py
00-kat: I'm fine with `* 60` but I didn't want to make them consts because `SECONDS_IN_TEN_MINUTES`/`SECONDS_IN_FIFTEEN_MINUTES` is very wordy (same with `10 * SECONDS_IN_MINUTE` in my opinion), and `TEN_MINUTES`/`FIFTEEN_MINUTES` doesn't make it obvious what that number even means if you don't check its value.

One thing I thought of initially but discarded since I felt a `ten_minutes = dt.timedelta(minutes=10).total_seconds()` was good enough is to make a `MonotonicTime` class (in toolbox) that wraps a float, but defines a subtraction operator that returns `timedelta` (since being unable to compare `timedelta`s with floats was the reason I decided to just use floats for everything). Should I do that instead?

(**Edit**: I previously thought the operators I needed to define were a total order when compared with timedelta, but I later realized I meant subtraction of `MonotonicTime`s.)

00-kat: Decided to do that in a separate commit: 63ca4489079fbaaba56aa2d0cc9a689517186e94 (the [full diff] is also useful). I'll squash it into this PR if you're fine with this approach.

**Edit**: I've also added tests (they were as painful to write as they are to read).

[full diff]: https://github.com/ghostty-org/discord-bot/compare/4e4d0fbd5f2827559671e0c8a9e0c6cf045cdcc0...63ca4489079fbaaba56aa2d0cc9a689517186e94ad9

CoderJoshDK: > I've also added tests (they were as painful to write as they are to read).

image

They sure are hard to read considering they don't exist .... ?

00-kat: I'm referring to 63ca4489079fbaaba56aa2d0cc9a689517186e94, not the current state of this PR.

trag1c: > […] `TEN_MINUTES`/`FIFTEEN_MINUTES` doesn't make it obvious what that number even means if you don't check its value.

So, nothing would change compared to the current state of the PR 😉 It's just that the definition is local instead of global. I don't think it's really an issue anyway.

The `MonotonicTime` approach is well done but feels overengineered to me.

00-kat: > It's just that the definition is local instead of global. I don't think it's really an issue anyway.

Everything in the function is concerned with seconds, so it's a bit more obvious in my opinion. (Apart from being closer to the use-site, if it's local it can't be used anywhere else in the future without thinking about what to name it.)

trag1c: I guess that's fair. Also now that it's int mult it's way better than the timedelta conversion eff-wise so I suppose say it can stay this way =)

trag1c (approved):

PR #555, opened 2026-05-26
Fix review request event merging

00-kat:
Follow-up to #541.

I manually tested it this time properly™ with the same dummy data as with #541; this time the code can be found at c52adec9fc76ecabda87952bec09f8ec19e7c28a.

From the commit message:
> I previously mistakenly read
https://github.com/trag1c/monalisten/blob/01e239f0d1217a940cb0c31b52727ab5b4328ac7/src/monalisten/_core.py#L180 to mean “run all hooks concurrently”, but upon closer inspection it actually simply runs all hooks *for the same event* concurrently. Currently, Monalisten runs everything synchronously w.r.t. the SSE server, which results in handle_review_request() blocking for five seconds before sending the exactly one request that was handled in that time. This change simply runs it in the background.
>
> While it might seem like Monalisten itself should be doing this for every hook, that would result in webhook events being sent out of order in some cases (though I'm not sure if this matters, since GitHub puts no effort into sending them in order anyway). More problematically, state shared between hooks, such as the vouch queue, would need to be protected, despite every current hook running fast *enough* (modulo the Discord API call) that concurrency wouldn't matter within the webhook.

### DIFF
--- a/app/components/github_integration/webhooks/integration.py
+++ b/app/components/github_integration/webhooks/integration.py
@@ -51,7 +51,7 @@ class GitHubWebhooks(commands.Cog):
             if (token := config().webhook.secret)
             else None,
         )
-        self._monalisten_task: asyncio.Task[None] | None = None
+        self._tasks = set[asyncio.Task[None]]()
         self._vouch_queue: VouchQueue = {}
 
     @override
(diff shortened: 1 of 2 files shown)


CoderJoshDK (approved):
This makes enough sense to me. Even if this doesn't solve it, it at least isn't worse.

trag1c (approved):

PR #558, opened 2026-06-14
add field descriptions to the CLI

### LINKED ISSUE #505: Add field descriptions to the CLI
Follow-up to #504. We can take some from `CONTRIBUTING.md`.

trag1c:
Closes #505

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -53,22 +53,28 @@ def _alias(name: str) -> AliasChoices:
 
 
 class ConfigRoles(BaseModel):
-    mod: int
-    helper: int
+    mod: Annotated[int, Field(description="the id of the mod role")]
+    helper: Annotated[int, Field(description="the id of the helper role")]
 
 
 class ConfigTokens(BaseModel):
-    discord: SecretStr
-    github: SecretStr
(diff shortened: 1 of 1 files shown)


app/config.py
```diff
@@ -78,15 +84,28 @@ class Channels(NamedTuple):
...
 
 
 class ConfigWebhook(BaseModel):
     _bot: dc.Client
-    url: SecretStr
-    secret: SecretStr | None = None
-    channel_ids: Annotated[WebhookChannels, Field(alias="channels")]
+    url: Annotated[SecretStr, Field(description="the URL to receive events from")]
+    secret: Annotated[
+        SecretStr | None, Field(description="a token for validating events (optional)")
```
00-kat: Do we need the “(optional)”s? Non-optional ones have a “(required)” already added by Pydantic.

00-kat (approved):

app/config.py
trag1c: *You're absolutely right! ✨* I would say yes to make it consistent *but* I noticed that Pydantic also already adds `(default value: ...)` which better implies being optional; will remove =)

00-kat (approved):

PR #561, opened 2026-06-26
chore: update dependencies

CoderJoshDK:
One thing to keep in mind is that (based)pyright now OOMs if you don't add `NODE_OPTIONS=--max-old-space-size=8192`. The exact number doesn't have to be `8192`. But that works.


trag1c (approved):
Looks good, thanks!

PR #566, opened 2026-08-18
entities: support gated mentions in private repos

### LINKED ISSUE #284: Support private org repositories
* Special-case `meta#12` and `gh-bot#3` mentions
* Change existing `bot#12` to `dc-bot#12` to disambiguate the two (feel free to name them something better than `gh-bot`/`dc-bot`)
* Support linking code, comments, and events from those repos
* Only let moderators and maintainers do all of the above

trag1c:
Closes #284 

The tests were 100% slopped but I smoothed out a few things and I think they're *fine*.


app/components/github_integration/repositories.py
```diff
@@ -0,0 +1,24 @@
...
+    """Whether the message author may make the bot expose data from this repo."""
+    if not is_private_repo(owner, repo):
+        return True
+
+    cfg = config()
+    member = cfg.ghostty_guild.get_member(author.id)
+    return member is not None and (
+        member.get_role(cfg.role_ids.mod) is not None
+        or member.get_role(cfg.role_ids.maintainer) is not None
+    )
```
00-kat: I don't like that this is very similar to these helper functions (since it's quite far from the similar code): https://github.com/ghostty-org/discord-bot/blob/29c0a16878b13ded06736cf20ad8d2c4e33f4cb6/app/config.py#L202-L210

Maybe we should have an `is_privileged_discord` (`is_privileged_on_discord` maybe?) that checks mod+helper, and an `is_privileged_github` that checks mod+maintainer?

(Also, why not inline `config()` (i.e.: `config().ghostty…`)? As far as I know that's what we usually do, and if we're being extra paranoid about mutation for this function, I don't think it could be concurrently changed if there are no `await`s.)

CONTRIBUTING.md
00-kat:
```suggestion
The following roles will be necessary (all requiring the Manage Messages
```

00-kat: That suggestion looks very wrong but the diff is “**both** requiring” → “**all** requiring”.

app/components/github_integration/repositories.py
trag1c: > Maybe we should have an `is_privileged_discord` (`is_privileged_on_discord` maybe?) that checks mod+helper, and an `is_privileged_github` that checks mod+maintainer?

Yeah I was considering that but didn't do it in the end, I guess you're the convincing comment lol. I'll do that!

> (Also, why not inline `config()` (i.e.: `config().ghostty…`)? As far as I know that's what we usually do, and if we're being extra paranoid about mutation for this function, I don't think it could be concurrently changed if there are no `await`s.)

I think it was mostly the three `config()` calls looking expensive in my head even though they're actually cheap lol, will change.

app/config.py
```diff
@@ -34,12 +34,16 @@
...
+    "meta": "meta",
     "web": "website",
     "website": "website",
     "discord-bot": "discord-bot",
     "bot": "discord-bot",
     "bobr": "discord-bot",
 }
+# Repositories in this set can only be linked by privileged Discord members. The full
+# owner/name is kept here so the restriction also applies to full GitHub URLs.
+PRIVATE_REPOS = frozenset({("ghostty-org", "meta")})
```
00-kat: Should we just assume all private repos are in `"ghostty-org"`?

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
+from types import SimpleNamespace
```
00-kat: These should probably all be `Mock`s no? cf. #512.

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import TYPE_CHECKING, cast
+from unittest.mock import AsyncMock, Mock
+
+import discord as dc
+import pytest
+
+from app.components.github_integration import repositories
+from app.components.github_integration.code_links import CodeLinks, ContentCache
+from app.components.github_integration.comments import fetching
```
00-kat: `fetching` feels too generic a name given we're also importing code links, commit links, and entities.

```suggestion
from app.components.github_integration.comments.fetching import (
    comment_cache,
    COMMENT_PATTERN,
    get_comments,
)
```

or maybe

```suggestion
from app.components.github_integration.comments import fetching as comments
```

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def mock_repo_access(
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    privileged: bool,
+) -> None:
+    member = Mock(dc.Member)
+    guild = Mock(dc.Guild)
+    guild.get_member.return_value = member
```
00-kat: Could this be:

```suggestion
    guild = Mock(dc.Guild, get_member=Mock((), return_value=member))
```

(I don't really mind, I just felt it might be slightly more consistent.)

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+        is_privileged_github=Mock((), return_value=privileged),
+    )
+    monkeypatch.setattr(repositories, "config", Mock((), return_value=cfg))
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize(
+    ("privileged", "expected"),
+    [
+        (True, [(("ghostty-org", "meta", 12), None)]),
```
00-kat: Should that `12` perhaps be generated by Hypothesis? I'm not too fussed but I don't like hardcoded meaningless constants lol.

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+async def test_meta_entity_mentions_require_privilege(
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    privileged: bool,
+    expected: list[tuple[tuple[str, str, int], None]],
+) -> None:
+    mock_repo_access(monkeypatch, privileged=privileged)
+
+    assert [
+        sig async for sig in resolve_entity_signatures(make_message("meta#12"))
```
00-kat: Maybe also `https://github.com/ghostty-org/meta/issues/12` and `ghostty-org/meta#12`.

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+    privileged: bool,
+) -> None:
+    member = Mock(dc.Member)
+    guild = Mock(dc.Guild)
+    guild.get_member.return_value = member
+    cfg = SimpleNamespace(
+        ghostty_guild=guild,
+        is_privileged_github=Mock((), return_value=privileged),
+    )
+    monkeypatch.setattr(repositories, "config", Mock((), return_value=cfg))
```
00-kat: Is there no way to simply set the context variable? (Ditto the other monkeypatch.)

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+
+@pytest.mark.anyio
+async def test_private_repo_links_are_filtered_before_fetching(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    mock_repo_access(monkeypatch, privileged=False)
+
+    code_links = CodeLinks.__new__(CodeLinks)
+    code_links.cache = Mock(ContentCache)
+    code_links.cache.get = AsyncMock(return_value="secret")
```
00-kat: Same as that other style (I think) comment.

```suggestion
    code_links.cache = Mock(ContentCache, get=AsyncMock(return_value="secret"))
```

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+    config.assert_not_called()
+
+
+@pytest.mark.anyio
+async def test_private_repo_links_are_filtered_before_fetching(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    mock_repo_access(monkeypatch, privileged=False)
+
+    code_links = CodeLinks.__new__(CodeLinks)
```
00-kat: Wait… what. Is there some reason this is better than `CodeLinks(Mock(GhosttyBot))`?

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+    monkeypatch.setattr(repositories, "config", config)
+
+    assert repositories.can_link_repo(
+        cast("Account", SimpleNamespace(id=1)), "ghostty-org", "ghostty"
+    )
+    config.assert_not_called()
+
+
+@pytest.mark.anyio
+async def test_private_repo_links_are_filtered_before_fetching(
```
00-kat: This seems to be three separate tests in one? I don't see any state shared between them (apart from that duplicable monkeypatch).

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+    code_message = make_message(
+        "https://github.com/ghostty-org/meta/blob/main/private.py#L1"
+    )
+    assert repositories.is_private_repo("ghostty-org", "meta")
+    assert [snippet async for snippet in code_links.get_snippets(code_message)] == []
+    code_links.cache.get.assert_not_awaited()
+
+    comment_get = AsyncMock()
+    monkeypatch.setattr(fetching.comment_cache, "get", comment_get)
+    for anchor in ("issuecomment-123", "event-456"):
```
00-kat: If you split this up into its own test: this seems like a manual `pytest.mark.parametrize`.

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+        comment_message = make_message(
+            f"https://github.com/ghostty-org/meta/issues/12#{anchor}"
+        )
+        assert fetching.COMMENT_PATTERN.search(comment_message.content)
```
00-kat: The regex being tested isolate of `get_comments` (its user) makes me feel queasy for some inexplicable reason… Should those regexes be exported anyway?

```suggestion
```

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,111 @@
...
+        assert [
+            comment async for comment in fetching.get_comments(comment_message)
+        ] == []
+    comment_get.assert_not_awaited()
+
+    message = make_message("https://github.com/ghostty-org/meta/commit/abcdef0")
+    matches = COMMIT_SHA_PATTERN.findall(message.content)
```
00-kat, line 104: Ditto. Should this be pulled out into a separate function to use here? https://github.com/ghostty-org/discord-bot/blob/29c0a16878b13ded06736cf20ad8d2c4e33f4cb6/app/components/github_integration/commit_links.py#L130-L131

app/config.py
trag1c: I guess so yeah, will drop this.

tests/test_private_github_repos.py
trag1c: Yes! I forgot already 😭

tests/test_private_github_repos.py
trag1c: Didn't even notice this (like I said a lot of this was slopped), thanks!

tests/test_private_github_repos.py
trag1c: Yeah good call.

tests/test_private_github_repos.py
trag1c: Eeeh I don't think we need to reach for it here.

tests/test_private_github_repos.py
trag1c: Idfk what I tried before but I was getting a weird forwardref evaluation error, maybe I mistyped something 💀 Will fix!

app/components/github_integration/repositories.py
```diff
@@ -0,0 +1,17 @@
...
+
+if TYPE_CHECKING:
+    from toolbox.discord import Account
+
+
+def can_link_repo(author: Account, owner: str, repo: str) -> bool:
+    """Whether the message author may make the bot expose data from this repo."""
+    if owner != "ghostty-org":
+        return True
+    if repo not in PRIVATE_REPOS:
```
00-kat:
```suggestion
    if owner.casefold() != "ghostty-org":
        return True
    if repo.casefold() not in PRIVATE_REPOS:
```

You dropped these in the last force-push.

tests/test_private_github_repos.py
trag1c: Mmm yeah I see where you're coming from but I think it's fine here since we're not *just* testing the regex itself but rather showing that we get zero comments despite the regex catching stuff.

tests/test_private_github_repos.py
trag1c: I'm not sure what you mean sorry.

tests/test_private_github_repos.py
00-kat: Fair enough; I guess I was just worried because `get_comments` might (after a future change) include extra logic that's hard to express in regex.

tests/test_private_github_repos.py
00-kat: Those two lines I linked is the part of `COMMIT_SHA_PATTERN`'s user that's duplicated here.

tests/test_private_github_repos.py
trag1c: There is lol, an agent did this and I just thought "sure if that's what we have to do". Just checked and a cvar just works; will fix, sorry!

tests/test_private_github_repos.py
00-kat: (You can minimize this; it's the same thing as https://github.com/ghostty-org/discord-bot/pull/566#discussion_r3808072091 just with a different regex.)

tests/test_private_github_repos.py
```diff
@@ -0,0 +1,109 @@
...
+    code_links.cache.get.assert_not_awaited()
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("anchor", ["issuecomment-123", "event-456"])
+async def test_private_event_links_are_filtered_before_fetching(anchor: str) -> None:
+    mock_repo_access(privileged=False)
+
+    comment_get = AsyncMock()
+    comment_cache.get = comment_get
```
00-kat: I know this technically isn't a problem here but this makes me feel 🥴 since nothing resets it.

trag1c: Oh yeah true; is the most recent force-push good?

00-kat: Functions doubling as namespaces will never stop feeling weird, but oh that's cool.

00-kat (approved):

00-kat (approved):

PR #568, opened 2026-09-06
hcb: don't die on timeout + other improvements

trag1c:
Fixes #having-to-`/reload`-every-few-days

The first three commits include some notes =)

The tests were drafted by GPT-6 Astra but edited quite a bit by me, the actual changes were done by myself.


app/components/hcb_feed.py
```diff
@@ -111,47 +123,87 @@ async def publish_transaction(self, txn: hcb.Transaction) -> None:
...
 
     @tasks.loop(minutes=3)
-    async def update_feed(self) -> None:
-        if self.file_lock.locked():
+    async def feed_loop(self) -> None:
+        try:
+            await self._update_feed()
+        except Exception:
+            self.poll_failed = True
+            logger.exception("HCB feed poll failed; retrying on next scheduled poll")
```
00-kat, line 134: Do we not want to log what the exception was?

trag1c: We do, and it's logged.

00-kat (approved):

PR #572, opened 2026-09-14
hcb: add a temporary log for better prod insights

trag1c:


app/components/hcb_feed.py
```diff
@@ -145,6 +145,14 @@ async def _update_feed(self) -> None:
 
         logger.debug("fetching HCB feed transactions")
         resp = await self.org.async_get_transactions(expand="donation", per_page=100)
+
+        # Temporary log for response and behavior tracking, to be removed soon hopefully
```
00-kat, line 149:
```suggestion
        # XXX: temporary log for response and behavior tracking, to be removed soon hopefully
```
maybe, for greppabiilty?

00-kat (approved):

app/components/hcb_feed.py
trag1c: Perchance…

trag1c: OTOH I do intend for this to be genuinely temporary, so I don't feel like it's that much needed

00-kat: Fair lol, idrc.

00-kat (approved):

PR #576, opened 2026-09-17
entities: support PR stacks

### LINKED ISSUE #575: Stack mentions
Stacks steal an entity ID: after #572 is #574, since #573 is the stack for #571+#572 (https://api.github.com/repos/ghostty-org/discord-bot/stacks/573). Currently, the bot sends nothing in response to a mention like `#573`, but it should send something about the stack.

trag1c:
Closes #575.


app/components/github_integration/entities/fmt.py
```diff
@@ -73,7 +80,42 @@ def _format_entity_detail(entity: Entity) -> str:
...
+    return [*start, placeholder, *end]
+
+
+def _format_pr_stack(stack: PRStack) -> str:
+    heading = f"{emojis()['stack']} **Stack #{stack.number}**"
+
+    prs = stack.pull_requests
+    pr_list = _collapse(
+        [
+            f"* {get_entity_emoji(pr)} {STACKED_PR_TEMPLATE.format(pr=pr)}"
```
00-kat: Please escape the title: https://github.com/ghostty-org/discord-bot/blob/6e207699847763b14df3c4514f046163574d4d21/app/components/github_integration/entities/fmt.py#L119

app/components/github_integration/entities/fmt.py
```diff
@@ -16,16 +22,17 @@
     from app.components.github_integration.models import Entity
 
 ENTITY_TEMPLATE = "**{entity.kind} [#{entity.number}](<{entity.html_url}>):** {title}"
+STACKED_PR_TEMPLATE = "**[#{pr.number}](<{pr.html_url}>):** {pr.title}"
```
00-kat: Why not just inline these?

app/components/github_integration/entities/fmt.py
```diff
@@ -73,7 +80,42 @@ def _format_entity_detail(entity: Entity) -> str:
...
+        [
+            f"* {get_entity_emoji(pr)} {STACKED_PR_TEMPLATE.format(pr=pr)}"
+            for pr in reversed(prs)
+        ],
+        limit=9,
+        placeholder=f"* … *({len(prs) - 9} more)*",
+    )
+
+    owner, name = stack.owner, stack.repo_name
+    base = f"[`{owner}/{name}:{prs[0].base_ref}`](<https://github.com/{owner}/{name}>)"
```
00-kat, line 104: …can you somehow break GitHub into having a stack with 0 PRs (by editing them out)? This index would error if so (as would everything else I guess lol).

app/components/github_integration/entities/fmt.py
```diff
@@ -73,7 +80,42 @@ def _format_entity_detail(entity: Entity) -> str:
...
+    path_parts = _collapse(
+        [base, *(f"`{branch}`" for branch in (pr.head_ref for pr in prs))],
```
00-kat: Why can't this just be:
```suggestion
        [base, *(f"`{pr.head_ref}`" for pr in prs],
```

It being that way is rather confusing.

app/components/github_integration/models.py
```diff
@@ -126,6 +126,22 @@ class Discussion(Entity, frozen=True):
     state_reason: Literal["DUPLICATE", "RESOLVED", "OUTDATED", "REOPENED"] | None
 
 
+class PRStack(Entity, frozen=True):
```
00-kat, line 131: Is it possible to set a default value on inherited properties? And if not, would you mind overriding `__init__` and supplying a default value for all the stuff that has a dummy value in `entities/cache.py`?

00-kat, line 131: This could use a comment as to why this inherits from `Entity`.

00-kat (changes requested):

app/components/github_integration/entities/fmt.py
trag1c: No idea, I guess we'll find out in production lmao

app/components/github_integration/entities/fmt.py
trag1c: It can defo be that way, there was some more stuff initially but it can indeed now be cleaned up =)

app/components/github_integration/entities/fmt.py
trag1c: I guess we can

app/components/github_integration/models.py
trag1c: Yeah it is possible and it's as simple as you'd think lmao

app/components/github_integration/entities/fmt.py
00-kat: Lmao fair.

app/components/github_integration/models.py
```diff
@@ -126,6 +126,27 @@ class Discussion(Entity, frozen=True):
     state_reason: Literal["DUPLICATE", "RESOLVED", "OUTDATED", "REOPENED"] | None
 
 
+# This inherits from Entity only to avoid special-casing the stack in all places that
+# rely on the entity cache (i.e. expect an Entity).
+class PRStack(Entity, frozen=True):
+    number: int
+    pull_requests: list[StackedPR]
+    title: str = ""
```
00-kat, line 135: This could also use a comment.

```suggestion
    # Fields of Entity that are meaningless on a PR stack.
    title: str = ""
```

trag1c: Indeed, thanks!

app/components/github_integration/models.py
```diff
@@ -126,6 +126,28 @@ class Discussion(Entity, frozen=True):
     state_reason: Literal["DUPLICATE", "RESOLVED", "OUTDATED", "REOPENED"] | None
 
 
+# This inherits from Entity only to avoid special-casing the stack in all places that
+# rely on the entity cache (i.e. expect an Entity).
+class PRStack(Entity, frozen=True):
+    number: int
+    pull_requests: list[StackedPR]
+    # Fields of Entity that are meaningless for a stack.
```
00-kat:
```suggestion
    # Fields of `Entity` that are meaningless for a stack.
```
Oh oops I forgot.

00-kat (dismissed):

app/components/github_integration/models.py
trag1c: Same lol, wanted to add them but an inner devil made me not.

00-kat (approved):

PR #578, opened 2026-09-18
hcb: store up to 500 transactions; include date in key

trag1c:


app/components/hcb_feed.py
```diff
@@ -168,20 +170,11 @@ async def _update_feed(self) -> None:
...
-        retained_ids = sent_ids & transactions.keys()
-        if retained_ids != sent_ids:
-            self._save_history(retained_ids)
-        sent_ids = retained_ids
-
-        new_ids = sorted(
-            transactions.keys() - sent_ids,
-            key=lambda txn_id: (date_sort_key(transactions[txn_id]), txn_id),
-        )
+        new_ids = sorted(transactions.keys() - sent_ids)
```
CoderJoshDK: Ahhhh. I see. This is smart

CoderJoshDK (dismissed):

CoderJoshDK (approved):
I think this is very decent looking. As a follow up though, I have one thought. We do a lot of re-writing the whole file over and over. Would it not be better to have the job append new keys and then at the end, cut the file down to size?

00-kat (approved):

00-kat:
> We do a lot of re-writing the whole file over and over. Would it not be better to have the job append new keys and then at the end, cut the file down to size?

Does it really matter for the amount of data we store? This is *far* easier to implement after all.

CoderJoshDK:
> Does it really matter for the amount of data we store? 

This isn't about data amount. This is about representation. We are under the ghostty umbrella. We should strive for optimized code even where it shouldn't matter. And besides, well optimized code is nice!

> This is far easier to implement after all.

I don't think it will be difficult. It should be pretty simple to implement.

trag1c:
> We are under the ghostty umbrella. We should strive for optimized code even where it shouldn't matter.

Are you suggesting we rewrite in Rust or Zig? :trollface:

00-kat:
> I don't think it will be difficult. It should be pretty simple to implement.

Objectively sure; I meant comparatively.

PR #580, opened 2026-09-18
Make the “failed to load emojis” log clearer

00-kat:
Re https://discord.com/channels/1281624934636064811/1281624935558807678/1550637622609842327.


CoderJoshDK (approved):
Ah. Yea, that makes more sense

PR #581, opened 2026-09-19
perf: reduce IO operations

CoderJoshDK:
Don't do a full file write and replace on every single transaction. Do a single write after all new transactions are complete.
Partial completion still works well because duplicates are ignored in the next run thanks to processing everything as a set.
This is the followup of https://github.com/ghostty-org/discord-bot/pull/578#pullrequestreview-5253508696

In theory, if all new transactions came in one at a time, this would be _more_ IO. But we observe that not to be the case. Many recurring transactions show up together. If we wanted to really overcomplicate this, we could skip the single append when there is only 1 new transaction. But that feels even more complex.


00-kat:
One thing I don't get: why did `_save_history` ever have to run for every transaction, instead of only at the end? I.e.: why do we need to append it before all the transactions are sent?

trag1c:
> One thing I don't get: why did `_save_history` ever have to run for every transaction, instead of only at the end? I.e.: why do we need to append it before all the transactions are sent?

So that if something goes wrong when publishing a transaction, the previous ones in the batch don't get resent on retry.

00-kat:
Then why not just write in a `finally:`?

trag1c:
> Then why not just write in a `finally:`?

Uhh it's a bit complicated so I wrote up a little thing: https://gist.github.com/trag1c/5aa5771eb822c568aea8a3e470e0a9ce

00-kat:
Lmao. @CoderJoshDK in that case what do you think about just writing once at the end in a `finally:` block?

CoderJoshDK:
I did make 73796396846039a83171dfa884629cc4fb7c685d use a finally. But if the publish is already in a try-catch block, this feels unneeded. I actually don't see a case where the new `try` block will catch anything.
If there is some IO failure, at least one published transaction will be re-sent. I guess the question is if you want to limit that to a single transaction or a batch of transactions. But then again, more IO operations increase your chance of IO failure. 

What I would actually like to do is:

```diff
diff --git a/app/components/hcb_feed.py b/app/components/hcb_feed.py
index d74fa8b..a6169a2 100644
--- a/app/components/hcb_feed.py
+++ b/app/components/hcb_feed.py
@@ -191,21 +191,18 @@ class HCBFeed(commands.Cog):
             txn_ids=", ".join(new_transactions),
         )
         updated = False
-        try:
-            for txn_key, txn in new_transactions.items():
-                try:
-                    if await self.publish_transaction(txn):
-                        sent_keys.add(txn_key)
-                        updated = True
-                except Exception:
-                    logger.exception(
-                        "failed to publish HCB transaction {txn_id!r}; leaving for "
-                        "retry",
-                        txn_id=txn_key,
-                    )
-        finally:
-            if updated:
-                self._save_history(sent_keys)
+        for txn_key, txn in new_transactions.items():
+            try:
+                if await self.publish_transaction(txn):
+                    sent_keys.add(txn_key)
+                    updated = True
+            except Exception:
+                logger.exception(
+                    "failed to publish HCB transaction {txn_id!r}; leaving for retry",
+                    txn_id=txn_key,
+                )
+        if updated:
+            self._save_history(sent_keys)

     def _save_history(self, transaction_ids: Iterable[str]) -> None:
         temp = self.history_file.with_suffix(".tmp")
```

CoderJoshDK:
Yea, I like that better. So I made that change.

00-kat:
> But if the publish is already in a try-catch block, this feels unneeded.

Yeah I noticed that too but the `finally:` would still catch `BaseException` so I wasn't sure.

00-kat (approved):

00-kat:
I thought we should wait for trag1c's review.

